PR #4, opened 2021-09-27
Fixed implementation for .merge()

Miestrode:
Merge currently doesn't handle end of files well. The new approach works with end of files as if they were concrete values.


Miestrode:
This is now rendered useless since the new LL(K) branch works by using internal indices in the stream.

PR #6, opened 2021-10-01
Update stream.rs

Miestrode:
End of files should be a physical, non zero-width range in the input. Since many error reporting libraries represent zero-width spans in a bad way.


Miestrode:
changing the method a bit, 1 second

PR #17, opened 2021-11-02
Added SeparatedBy::at_least method

oOBoomberOo:
allows you to specify the minimum amount of items needed to match inside `Parser::separated_by()`.

I've also added test cases to make sure this work with `SeparatedBy::allow_trailing` and `SeparatedBy::allow_leading`

### DIFF
--- a/src/combinator.rs
+++ b/src/combinator.rs
@@ -659,6 +659,15 @@ mod tests {
     }
 
     #[test]
+    fn separated_by_at_least_without_trailing() {
+        let parser = just::<_, Simple<char>>('-')
+            .separated_by(just('.'))
+            .at_least(3);
+        
+        assert!(parser.parse("-.-.-.").is_err());
+    }
+
+    #[test]
     fn separated_by_at_least_with_leading() {
         let parser = just::<_, Simple<char>>('-')
             .separated_by(just('.'))
--- a/src/combinator.rs
+++ b/src/combinator.rs
@@ -659,6 +659,15 @@ mod tests {
     }
 
     #[test]
+    fn separated_by_at_least_without_leading() {
+        let parser = just::<_, Simple<char>>('-')
+            .separated_by(just('.'))
+            .at_least(3);
+        
+        assert!(parser.parse(".-.-.-").is_err());
+    }
+
+    #[test]
     fn separated_by_at_least_without_trailing() {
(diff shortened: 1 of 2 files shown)


src/combinator.rs
```diff
@@ -625,3 +642,41 @@ impl<I: Clone, O, A: Parser<I, O, Error = E>, U: Clone, E: Error<I>> Parser<I, U
...
+
+    use super::*;
+
+    #[test]
+    fn separated_by_at_least() {
+        let parser = just::<_, Simple<char>>('-')
+            .separated_by(just('.'))
+            .at_least(3);
+        
+        assert_eq!(parser.parse("-.-.-"), Ok(vec!['-', '-', '-']));
```
zesterer, line 658: Perhaps an error-generating pattern here too like `-.-.`? For me, it's quite important that we make sure that this continues to correctly interpret trailing/leading patterns as errors when used as `at_least` and without `allow_trailing`/`allow_leading`.

src/combinator.rs
```diff
@@ -625,3 +642,41 @@ impl<I: Clone, O, A: Parser<I, O, Error = E>, U: Clone, E: Error<I>> Parser<I, U
     #[inline]
     fn parse_inner_silent(&self, d: &mut Silent, s: &mut StreamOf<I, E>) -> PResult<I, U, E> { #[allow(deprecated)] self.parse_inner(d, s) }
 }
+
+#[cfg(test)]
+mod tests {
```
zesterer, line 647: I appreciate the tests. This combinator is by far the most complicated and I often have a hard time wrapping my head around the logic going on.

zesterer:
Thanks for the PR, I really appreciate this change! Perhaps a similar example could be added to `Repeated::at_least`?

src/combinator.rs
oOBoomberOo: Good point, I'll add that.

oOBoomberOo:
As for `Repeated::at_least` I believe that can be added in its own PR (although I would prefer to apply `rustfmt` to the project first)

zesterer:
Merging, thanks very much. Regarding `rustfmt`: I'd like to make a few adjustments to `rustfmt`'s configuration before doing this because I've found it to do rather bizarre things to more complex generics. At the very least, I think that most trait bounds should be moved to `where` clauses before doing this.

PR #33, opened 2021-11-24
Rewind parser

ryo33:
Hi, continuously I perfectly enjoying parsing with this amazing crate. Today I introduce `lookahead` combinator.
I don't believe that this should be merged, but this is a little bit convenience for some case. Also, I am not confident on the naming and design of this new combinator.

## Motivation

I want to parse something like this;
```
a b c - d
```
with
```
word
    .repeated()
    .then(word
        .then_ignore(just('-'))
        .then(word))
```

But this fails at `-`.

So I introduce `lookahead` and change the parser like below.

```
word
    .then_ignore(none_of("-".chars()).lookahead())
    .repeated()
    .then(word
        .then_ignore(just('-'))
        .then(word))
```

## Compare to other designs

I choose the one I think it's most fits to chumsky's APIs.

### `word.lookahead(none_of("-".chars()))`

This is like very natural, but always ignores the outputs of the lookahead parser.
In some case, we may want not to ignore, such as infix attribute syntax that have influence to the both side.

### `word.then_ignore(lookahead(none_of("-".chars())))`

This is also natural, and allows us to choose whether ignore the output or not.
However, this is odd in chumsky's conventions.

### DIFF
--- a/src/combinator.rs
+++ b/src/combinator.rs
@@ -1022,6 +1022,55 @@ impl<I: Clone, O, A: Parser<I, O, Error = E>, U: Clone, E: Error<I>> Parser<I, U
     }
 }
 
+/// See [`Parser::lookahead`].
+pub struct Lookahead<A>(pub(crate) A);
+
+impl<I: Clone, O, E: Error<I>, A> Parser<I, O> for Lookahead<A>
+where
+    A: Parser<I, O, Error = E>,
+{
+    type Error = E;
+
+    fn parse_inner<D: Debugger>(
+        &self,
+        debugger: &mut D,
+        stream: &mut StreamOf<I, Self::Error>,
+    ) -> PResult<I, O, Self::Error>
+    where
+        Self: Sized,
+    {
+        let lookahead_from = stream.save();
+        match {
+            #[allow(deprecated)]
+            debugger.invoke(&self.0, stream)
+        } {
+            (errors, Ok((out, alt))) => {
+                stream.revert(lookahead_from);
+                (errors, Ok((out, alt)))
+            }
+            (errors, Err(err)) => (errors, Err(err)),
+        }
+    }
+
+    fn parse_inner_verbose(
+        &self,
+        d: &mut Verbose,
+        s: &mut StreamOf<I, Self::Error>,
+    ) -> PResult<I, O, Self::Error> {
+        #[allow(deprecated)]
+        self.parse_inner(d, s)
+    }
+
+    fn parse_inner_silent(
+        &self,
+        d: &mut Silent,
+        s: &mut StreamOf<I, Self::Error>,
+    ) -> PResult<I, O, Self::Error> {
+        #[allow(deprecated)]
+        self.parse_inner(d, s)
+    }
+}
+
 #[cfg(test)]
 mod tests {
     use error::Simple;
--- a/src/combinator.rs
+++ b/src/combinator.rs
@@ -1134,4 +1134,17 @@ mod tests {
             .chain(just(','));
         assert_eq!(parser.parse("-,-,-,"), Ok(vec!['-', '-', '-', ',']))
     }
+
+    #[test]
+    // TODO: this fails
+    fn reproduce_repeated_with_infix() {
+        let a = just::<_, Simple<char>>('a');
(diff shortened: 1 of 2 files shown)


ryo33:
@zesterer Thanks for the review.

> This doesn't look like a pattern that should fail, unless word also accepted - as an input. Is this the case?

I forgot to write that the `word` parser only parses alphanumerics.

> The PR seems like the parse succeeds on valid input, but reverts to the start of this input. Is this the intended behavior?

Yes, this is intended.

Certainly, It looks like it could be a success, and the introduced parser is unnecessary.
I may do something wrong or run into a pitfall. I'm going to reproduce the condition accurately.

src/combinator.rs
```diff
@@ -1085,4 +1134,17 @@ mod tests {
...
+        let a = just::<_, Simple<char>>('a');
+        let parser = a
+            .clone()
+            .repeated()
+            .then(a.clone().then_ignore(just('-')).then(a))
+            .then_ignore(end());
+
+        assert_eq!(parser.parse("aaa-a"), Ok((vec!['a', 'a'], ('a', 'a'))));
+    }
 }
```
ryo33, line 1150: I added that reproduces the condition I met and I want to solve.

zesterer:
What I'm worried about is the extent to which this combinator generalises and the ergonomics of it. For me, it took me quite a while to understand exactly what the code in your example was actually doing.

I think it's probably the case that this case of overzealous parsing is only hit when a parser consumes too much of the input to the right of its intended pattern. If this is the case, may I recommend a rename to `with_stop` and with a slight change in logic?

```
word
    .with_stop(word)
    .repeated()
    .then(word.then_ignore(just('-')).then(word))
```

ryo33:
I had a little survey on this, the `lookahead` combinator is a common use-case, I think. 
Parsec has a `lookahead` combinator, so combine crate also has it (combine refers to parsec on its API).
There is also a similar combinator `try`. Contrary to `lookahead`, `try` make a parser not to consume inputs when failed.
Therefore all chumsky parsers are already `try`. But there are no combinators that make a parser not consume inputs even when successful.

I roughly agree the lookahead API I suggested is difficult to understand. How about `parser.rewind()` or `parser.peek()`
I think `parser.lookahead()` is intuitive if get used to, like `parser.ignored()`. Also, I prefer `lookahead` because it is a common term.

zesterer:
`rewind` seems like a suitable name to me. We'll have to make sure that the documentation for these functions is good because their use isn't particularly intuitive.

> Also, How about word.not_followed_by(just('-'))

We already have `.then_ignore(..)`, so how about `.then_not(..)`?

src/lib.rs
```diff
@@ -953,6 +953,29 @@ pub trait Parser<I: Clone, O> {
...
+    /// ```
+    /// # use chumsky::prelude::*;
+    /// let just_numbers = text::digits::<_, Simple<char>>(10)
+    ///     .padded()
+    ///     .then_ignore(none_of("+-*/".chars()).rewind())
+    ///     .separated_by(just(','));
+    /// // 3 is not parsed because it's followed by '+'.
+    /// assert_eq!(just_numbers.parse("1, 2, 3 + 4"), Ok(vec!["1".to_string(), "2".to_string()]));
+    /// ```
+    fn rewind(self) -> Rewind<Self>
```
ryo33, line 972: I renamed it to `rewind`, and made the notification slightly better.

src/combinator.rs
```diff
@@ -1134,17 +1134,4 @@ mod tests {
             .chain(just(','));
         assert_eq!(parser.parse("-,-,-,"), Ok(vec!['-', '-', '-', ',']))
     }
-
-    #[test]
-    // TODO: this fails
-    fn reproduce_repeated_with_infix() {
```
zesterer: Does this need a `#[should_fail]`?

ryo33: Yes, but I deleted the test in my latest commit.

zesterer: Apparently not.

zesterer:
Thanks for your work on this PR!

ryo33:
Thank you! This makes my programming language more powerful 💪

Issue #45, opened 2021-12-06
Add a State parameter to parsers


kythyria:
In some cases, it's useful to be able to pass something around to parsers, such as a tree builder, interner, arena, or similar gizmo. Something that can't just be created ex nihilo in a leaf and passed outward (an `ego-tree` document, in my case: the parent of a node has to exist before the children do).

> zesterer: An extra type parameter on Parser with a default could work
> zesterer: trait Parser<I, O, State = ()>
> zesterer: Then an extra combinator like .map_with_state(|output, state: &mut State|)
> zesterer: And an extra Parser::parse_with_state function

Which is basically what Logos does modulo exactly how the state is retrieved.

zesterer:
So I've begun adding this in #46, but it's sadly resulting in a lot of type inference problems. I've yet to determine whether these are inherent problems or just circumstantial though.

zesterer:
I think I can get rid of the inference error with extra parameterisation...

PR #46, opened 2021-12-06
Began adding state param

zesterer:
As per #45


zesterer:
This is being implemented in #82, so this PR is now redundant.

PR #49, opened 2021-12-08
Make clippy happy

CraftSpider:
Run clippy on the codebase, fixed all but one or two warnings. Did each fix as its own commit, if you want to only pick some I'll scrap the rest, or if you don't want any I can close this. If you want a minimal set, I'd at least bench `redundant_clone` and `needless_range_loop`, as they *might* have an optimization impact. And `deprecated_semver` looks like an actual fix, if a relatively insignificant one.

### DIFF
--- a/src/chain.rs
+++ b/src/chain.rs
@@ -1,6 +1,4 @@
 
-use super::*;
-
 mod private {
     pub trait Sealed<T> {}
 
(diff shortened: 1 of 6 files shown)


zesterer (changes requested):
I'm broadly happy with these changes, although there are a few details I'd like to see accounted for before merge (as mentioned in other comments). Thanks!

zesterer:
> Sounds good! Question: Would you like me to add `#![allow()]` on relevant places, add comments, or just leave the warnings as-is in those places?

`#[allow(...)]` in those specific cases would be good, although I'm a little reluctant to enable clippy for this codebase because it's doing enough weird stuff that it's likely not going to be able to usefully reason about much of the code.

CraftSpider:
Added allow on all the locations, and tried replacing the import with a markdown link.

zesterer:
Thanks. Happy to merge?

zesterer:
Thanks for the PR!

PR #54, opened 2021-12-13
Add then_with combinator

CraftSpider:
Includes the two notes I thought important: That `or` is preferred if possible, and the warning about suboptimal errors. The `ThenWith` type requires the extra generics due to issues on the impl otherwise regarding unbound generics.

### DIFF
--- a/src/combinator.rs
+++ b/src/combinator.rs
@@ -229,6 +229,63 @@ impl<I: Clone, O, U, A: Parser<I, O, Error = E>, B: Parser<I, U, Error = E>, E: 
     }
 }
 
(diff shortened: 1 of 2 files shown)


CraftSpider:
Note: Could also change the resulting parser to `(I, U)` to match `then`, if we passed a reference to the parsed value (perhaps a mutable one) instead of ownership. Not sure which would be preferred.

zesterer:
Hopefully I will have time to review this tonight.

zesterer:
Narrator: *He did not...*

Tonight, then.

zesterer:
Thanks, this looks good!

PR #61, opened 2022-01-01
Make delimited_by take parsers as arguments

### LINKED ISSUE #52: `.delimited_by()` should take parsers as arguments, not input tokens
This is another inconsistency I noticed when working with the API. `.separated_by()` takes a parser as a separator, so it makes sense to expect that `.delimited_by()` take parsers as delimiters.

This would for example enable `.delimited_by()` to express surrounding XML tags in `char` parsers, which consist of multiple characters, and can contain parsable attributes themselves.

wackbyte:
Closes #52

### DIFF
--- a/src/combinator.rs
+++ b/src/combinator.rs
@@ -297,10 +297,15 @@ impl<I: Clone, O1, O2, A: Parser<I, O1, Error = E>, B: Parser<I, O2, Error = E>,
 
 /// See [`Parser::delimited_by`].
 #[derive(Copy, Clone)]
-pub struct DelimitedBy<A, I>(pub(crate) A, pub(crate) I, pub(crate) I);
+pub struct DelimitedBy<A, L, R, U> {
+    pub(crate) item: A,
+    pub(crate) start: L,
+    pub(crate) end: R,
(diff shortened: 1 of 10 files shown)


src/combinator.rs
```diff
@@ -297,10 +297,16 @@ impl<I: Clone, O1, O2, A: Parser<I, O1, Error = E>, B: Parser<I, O2, Error = E>,
 
 /// See [`Parser::delimited_by`].
 #[derive(Copy, Clone)]
-pub struct DelimitedBy<A, I>(pub(crate) A, pub(crate) I, pub(crate) I);
+pub struct DelimitedBy<A, L, R, U, V> {
+    pub(crate) item: A,
+    pub(crate) start: L,
+    pub(crate) end: R,
+    pub(crate) phantom_start: PhantomData<U>,
+    pub(crate) phantom_end: PhantomData<V>,
```
zesterer, line 305: Only a single `PhantomData` field is ever needed: if multiple type parameters are phantoms, you can use `phantom: PhantomData<(U, V)>`

zesterer:
Thanks for the PR! There's just one thing I'd like to see changed before merging.

PR #64, opened 2022-01-04
Implement support for `no_std`

### LINKED ISSUE #38: `no_std` support
Hey!

I've taken a look at the code and their doesn't seem to be any reason besides `error.rs` to make the crate `no_std` compatible (only taking the code into consideration). The problems with the above mentioned file are:
1) The error implements `std::error::Error`. This can be put behind a feature flag.
2) It uses `HashSet`, which is not available on `no_std`. The next best solution would be `BHashSet`, but that requires it's data to be `Ord`. And this doesn't always make sense.

Is this something you have considered/thought about? 

I am asking because I have a parser and wanted to have a version that runs on the web as well. `no_std` would be the best solution! :)

wackbyte:
I added a `std` feature to support building with `no_std`.

If enabled, it implements the `Error` trait and uses `std`'s `HashSet`.
Otherwise, it doesn't implement `Error` and uses `hashbrown`'s `HashSet`.

However, there are a few details I'd like to hear your opinion on if I need to make some changes.

* `no_std` is disabled in documentation.
* Because of rust-lang/cargo#1839, the dependency on `hashbrown` can't be optional. It could probably be if there was a `no_std` feature instead, but [according to the feature reference in the Cargo book](https://doc.rust-lang.org/cargo/reference/features.html#feature-unification), features should be additive, and a `std` feature should be used over a `no_std` one.
* I looked into `hashbrown` and found that it uses `ahash` by default (through a default feature), so I removed the direct dependency on it and used it through `hashbrown`. This may be undesirable.
* The `Verbose` struct in the `debug` module currently uses the `print!` macros. Since those do not exist in a `no_std` environment, the function that uses them was changed to be a no-op when `std` is disabled.

(Closes #38)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,13 +14,15 @@ exclude = [
 ]
 
 [features]
-default = ["ahash"]
+default = ["std"]
+# Integrate with the standard library
+std = []
 # Enable nightly-only features like better compiler diagnostics
 nightly = []
 
 [dependencies]
-# Use `ahash` for maintaining sets of expected inputs
-ahash = { version = "^0.3", optional = true }
+# Due to https://github.com/rust-lang/cargo/issues/1839, this can't be optional
+hashbrown = "0.11"
 
 [dev-dependencies]
 ariadne = "0.1.2"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,6 +16,7 @@ exclude = [
 [features]
 default = ["ahash", "std"]
 # Use `ahash` instead of the standard hasher for maintaining sets of expected inputs
+# (Also used if `std` is disabled)
 ahash = []
 # Integrate with the standard library
 std = []
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,13 +14,17 @@ exclude = [
 ]
 
 [features]
-default = ["std"]
+default = ["ahash", "std"]
+# Use `ahash` instead of the standard hasher for maintaining sets of expected inputs
+ahash = []
 # Integrate with the standard library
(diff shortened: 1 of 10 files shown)


zesterer:
I think, as you mention in the post, that this might be the wrong way to go about it. Cargo features are strictly additive, and so there should be an `std` feature as part of the `default-features` list instead. This should also solve the issue you encountered with `hashbrown`, I believe.

What do you think about this?

wackbyte:
I did use a `std` feature and added it to `default-features`, but the issue is that I can't enable an optional dependency if a feature is disabled.

zesterer:
This might require `hashbrown` being non-optional in this case, which is a shame. I'd prefer that over a 'negative' feature flag though, I think.

zesterer:
Hey, this has a few conflicts. Could you resolve them? Happy to merge after that.

zesterer:
Thanks very much! I'll probably try to find time to add a `no_std` target to the CI tomorrow (some nebulous embedded ARM platform will do) so we can make sure that nothing that relies on `std` accidentally sneaks in without the appropriate flag.

mainrs:
Glad you could work it out @wackbyte. I didn't get the hasher to compile properly. Your `ahash` feature and implementation was missing in my PoC. 

I should have updated the issue with my progress and problem though, maybe it could have been solved that way!
Thanks for the PR. `chumksy` on `wasm` is something I have been looking forward to!

zesterer:
Great to hear that the library fits this use-case. I'm interested to hear how it works out!

PR #75, opened 2022-01-16
Replace error `Vec` with a linked list of slab `Vec`s

domenicquirl:
You mentioned that you wanted to take some stabs at performance. While I am not familiar enough with the codebase to suggest any _structural_ changes (and sadly don't have the time to change that, as interesting as `chumsky` is), I had a look at the benchmarks and also `perf`ed them. The flamegraph has the usual problem of recursive and/or deeply layered parsing function that any parser expectedly consists mostly of its sub-parsers and it's hard to make out problematic areas. One part that showed up repeatedly was `Vec::push` and `Vec::append`. 

If I saw correctly, some parsers like `Seq` use a `Vec` as output as well, but on top-level I noticed all of the parsing errors are returned in a `Vec`, so many of the parsers copy them from outputs of their constituents into an aggregated result `Vec`. I replaced that with a wrapper around `LinkedList<Vec<T>>`  to obtain the `O(1)` characteristics of its `append`. It doesn't have everything that `Vec` has (in particular, no methods to remove elements), but everything that you use in `chumsky` plus `iter` and `into_iter`, so it's possible to get back to a `Vec` after parsing . On the Linux machine I tested this, making this change through all the parsers and also the recovery `Strategy`s made about a 10% difference. This depends heavily on the presence of errors, obviously.

Notables:
 - Breaking: the container type is exposed through `PResult`
 - Right now this PR increases the MSRV to current stable. That's only because I used `unwrap_unchecked` though, so if you want to support earlier versions that's an easy change
 - I added tests for the data structure, and also documented it briefly. I put the thing in its own new module, which is missing the quote in its module comment because I didn't feel qualified ^^

### DIFF
--- a/src/combinator.rs
+++ b/src/combinator.rs
@@ -162,7 +162,7 @@ impl<I: Clone, O, A: Parser<I, O, Error = E>, E: Error<I>> Parser<I, Option<O>> 
             debugger.invoke(&self.0, stream)
         }) {
             (errors, Ok((out, alt))) => (errors, Ok((Some(out), alt))),
-            (_, Err(err)) => (Vec::new(), Ok((None, Some(err)))),
+            (_, Err(err)) => (FlatList::new(), Ok((None, Some(err)))),
         }
     }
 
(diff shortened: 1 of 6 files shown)


zesterer:
Hey, thanks for this! I'm probably not going to merge this now because I'm working on a pretty big overhaul in #82, but I'll definitely be using this PR as a reference!

zesterer:
Thanks, this is very valuable information!

zesterer:
I'm going to close this because it's now out of date. However, we [did end up using](https://github.com/zesterer/chumsky/blob/main/src/input.rs#L709-L712) a very similar approach (and more besides), which was a big factor on the rewrite's very competitive performance. Thanks for the ideas!

PR #82, opened 2022-02-08
Add support for zero-copy parsing

### LINKED ISSUE #9: Support for zero-copy parsing
This is now being developed in #82

zesterer:
Currently, only some initial experiments outside of the main codebase.

The new design departs quite a lot from `master`, to the point that it's likely to end up being a near-complete rewrite of the crate. However, benchmarks so far are promising!

image

These benchmarks have a few caveats:

- `serde_json` is not quite comparable because it's not performing zero-copy parsing: the final `JsonValue` contains `String`.
- `pest` is not quite comparable because it does not support parsing byte strings: it's parsing a version of the input that's been run through `std::str::from_utf8`

Regardless, the benchmark demonstrates well the substantial performance improvement over `master` (old `chumsky` has been included at the bottom as a point of comparison).

Along with support for zero-copy parsing, this new design also permits the following:

- A state type parameter, allowing mutable state to be passed down through the parser (useful for interning and more)
- A check-only mode that skips generating output
- Regex parsers

Unfortunately, the code requires [GATs](https://github.com/rust-lang/rust/issues/44265), a feature that is currently unstable (but might not be for much longer!).

- [ ] Reimplement all primitives
- [ ] Reimplement all combinators
- [ ] Reimplement error prioritisation
- [ ] Reimplement recovery
- [ ] Reimplement all text parsers
- [ ] Remove `then_with`
- [ ] Switch default behaviour of `.parse(..)` to `.then_ignore(end())`, add `skip_all()` combinator so the old behaviour can be opted into with `.then_ignore(skip_all())`

Closes #9


oovm:
I want to experience this function but the following error is reported, how should I solve it?

```rust
error[E0309]: the parameter type `S` may not live long enough
   --> src\input.rs:233:64
    |
233 | ... E: Error<I::Token>, S, P: Parser<'a, I, E, S> + ?Sized>(parser: &P, inp: &mut InputRef<'a, '_, I, S>) -> PResult<Self, P::Output, E> {
    |                         -     ^^^^^^^^^^^^^^^^^^^ ...so that the type `S` will meet its required lifetime bounds...
    |                         |
    |                         help: consider adding an explicit lifetime bound...: `S: 'a`
    |
note: ...that is required by this bound
   --> src\input.rs:238:69
    |
238 | pub trait Parser<'a, I: Input + ?Sized, E: Error<I::Token> = (), S: 'a = ()> {
    |                                                                     ^^
error: missing required bound on `Iter`
   --> src\input.rs:402:5
    |
402 |     type Iter<'a>: Iterator<Item = T>;
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^-
    |                                      |
    |                                      help: add the required where clause: `where Self: 'a`
    |
    = note: this bound is currently required to ensure that impls have maximum flexibility
    = note: we are soliciting feedback, see issue #87479 <https://github.com/rust-lang/rust/issues/87479> for more information
```

zesterer:
> When zero-copy is merged, is it intended to fully replace the non-zero-copy code? If so, do you think it's a good time to remove the old code in this branch, lower the modules to root, and move the documentation over to the new modules?

Yes, I think this would be reasonable. I'm still not certain about the overall design though (the `Input` trait still feels like it has odd corner cases, we don't have error recovery systems (interesting discussion in #159 about some potential alternative strategies), and I think revisiting some of the API features like `from_nested` to make them more of a happy path is worth doing). I definitely don't see much of - if any - the old code surviving though.

toothbrush7777777:
Rust 1.65 will be [released tomorrow (3 November)](https://blog.rust-lang.org/inside-rust/2022/10/31/1.65.0-prerelease.html) and [supports generic associated types (GATs)](https://blog.rust-lang.org/2022/10/28/gats-stabilization.html).

PR #90, opened 2022-02-25
Add more combinators, fix generics on To

CraftSpider:
Adds a number of the missing combinators, including `ignore_then`/`then_ignore` and `fold{l,r}`. Also fixes some generics on certain impls.

### DIFF
--- a/src/input.rs
+++ b/src/input.rs
@@ -292,6 +292,13 @@ pub trait Parser<'a, I: Input + ?Sized, E: Error<I::Token> = (), S: 'a = ()> {
         MapWithSpan { parser: self, mapper: f }
     }
 
+    fn try_map<O, F: Fn(Self::Output, I::Span) -> Result<O, E>>(self, f: F) -> TryMap<Self, F>
+    where
+        Self: Sized,
+    {
+        TryMap { parser: self, mapper: f }
+    }
+
     fn ignored(self) -> Ignored<Self, E, S>
     where
         Self: Sized,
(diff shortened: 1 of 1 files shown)


src/input.rs
```diff
@@ -578,11 +706,11 @@ impl<F: Clone, I: ?Sized> Clone for Filter<F, I> {
     }
 }
 
-pub fn filter<F: Fn(&I::Token) -> bool, I: Input + ?Sized>(filter: F) -> Filter<F, I> {
+pub fn filter<F: Fn(&I::Token) -> bool, I: Input + ?Sized, E: Error<I::Token>>(filter: F) -> Filter<F, I, E> {
```
CraftSpider, line 709: Could also add the state param, if you want. I don't use it so I'm not sure if it's relevant here.

src/input.rs
```diff
@@ -373,6 +421,26 @@ macro_rules! go_extra {
     };
 }
 
+impl<'a, 'b, T, I, E, S> Parser<'a, I, E, S> for &'b T
```
CraftSpider, line 424: These two lifetimes were leftover from me trying something, I don't *think* they're necessary and can remove them if you want.

src/input.rs
```diff
@@ -1081,6 +1311,83 @@ where
...
+    A: IntoIterator,
+    A::IntoIter: DoubleEndedIterator,
+    F: Fn(A::Item, B) -> B,
+{
+    type Output = B;
+
+    fn go<M: Mode>(&self, inp: &mut InputRef<'a, '_, I, S>) -> PResult<M, Self::Output, E> where Self: Sized {
+        self.parser.go::<M>(inp)
+            .map(|out|
+                M::map(out, |(init, end)| init.into_iter().rev().fold(end, |b, a| (self.folder)(a, b)))
```
zesterer, line 1346: I wonder whether a dedicated combinator for this would be better eventually, avoiding the extra `Vec`. Something like `lhs.fold(rhs, |lhs. rhs| ...)` and `lhs.fold_into(rhs, |lhs, rhs| ...)`...

zesterer:
Thanks for the PR! All looks good to me.

PR #91, opened 2022-02-25
[zero-copy] Fix self-referential recursive

CraftSpider:
This isn't really original code, just copying over the way it works on master. My benchmarks no longer get OOM-killed with this change.


zesterer:
Thanks very much!

PR #93, opened 2022-02-27
[zero-copy] Fix duplicate inline attributes and a doc warning

wackbyte:
I merged the duplicate `#[inline]` and `#[inline(always)]` attributes on some functions into just `#[inline(always)]` and ported a doc fix from master.


zesterer:
Thanks!

zesterer:
Actually, should this branch be merging into `master` instead?

wackbyte:
Nope, since the duplicate attributes were only added in zero-copy and the docs were already fixed in master.

zesterer:
Oh, interesting. I don't remember doing this, I wonder why I did...

Issue #89, opened 2022-02-23
Add benchmarks against nom


knarkzel:
See https://github.com/Geal/nom/blob/main/examples/json.rs

zesterer:
I'll hopefully be doing this when #82 is ready for merge. Thanks for the reminder!

zesterer:
Closing this because it's been implemented in #94 for the `zero-copy` branch.

PR #94, opened 2022-02-28
[zero-copy] Add nom benchmark

wackbyte:
Adds a benchmark for nom based on the example linked in #89.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,4 +27,5 @@ regex = { version = "1.5", optional = true }
 [dev-dependencies]
 ariadne = "0.1.2"
 pom = "3.0"
+nom = "7.1"
 serde_json = { version = "1.0", features = ["preserve_order"] }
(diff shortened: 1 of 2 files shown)


wackbyte:
(here are the results I got on my Ryzen 5 3600X)
image

zesterer:
Holy heck! `chumsky/zero-copy` zero is *faster* than `nom`?! That I did not expect in the slightest. Obviously that's going to take a bit of a beating when proper error reporting is implemented, but hopefully most of the performance wins can be preserved.

zesterer:
Holy heck! `chumsky/zero-copy` zero is *faster* than `nom`, even without an explicit `escaped` primitive?! That I did not expect in the slightest. Obviously that's going to take a bit of a beating when proper error reporting is implemented (although the error handling the branch has right now is still roughly equivalent to `nom`'s), but hopefully most of the performance wins can be preserved.

zesterer:
As a confirmation, I'm seeing similar numbers locally.

image

PR #99, opened 2022-03-06
Support unicode character escapes in the JSON example

### LINKED ISSUE #86: Incorrect JSON parser example
I try to use the JSON parser under examples dir , mostly It works fine, but when I try to parse such snippet
```
{
 "test": "\u2222"
}
```
parser in example report a error, but actually this json snippet is valid according to the specification https://www.json.org/json-en.html

wackbyte:
The error messages are kind of strange, but it's functional

Closes #86


zesterer:
Thanks, this is great!

PR #102, opened 2022-03-12
Added optional `span:ident` argument to `select!`

### LINKED ISSUE #101: Support mapping with span in `select!`
Support this kind of syntax:

```rust
select! { span,
    Token::Word(w) => AstNode { span, ast: AstKind::Word(w) }
    Token::KeyWord(kw) => AstNode { span, ast: AstKind::KeyWord(kw) }
}
```

JohnDowson:
Closes #101

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1399,6 +1399,12 @@ impl<'a, I: Clone, O, E: Error<I>> Parser<I, O> for BoxedParser<'a, I, O, E> {
 /// ```
 #[macro_export]
 macro_rules! select {
+    ($span:ident, $($p:pat $(if $guard:expr)? => $out:expr),+ $(,)?) => ({
+        $crate::primitive::filter_map(move |$span, x| match x {
+            $($p $(if $guard)? => ::core::result::Result::Ok($out)),+,
+            _ => ::core::result::Result::Err($crate::error::Error::expected_input_found($span, ::core::option::Option::None, ::core::option::Option::Some(x))),
+        })
+    });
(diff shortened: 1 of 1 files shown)


zesterer:
I think I would prefer if this were implemented as an optional attribute that bound for each pattern, like so:

```rs
select! {
    Foo(foo), span => f(foo, span),
    Bar(bar) => bar,
}
```

This feels a lot cleaner to me. Are you okay if I reimplement this?

JohnDowson:
> This feels a lot cleaner to me.

While it does look cleaner, i'm not sure if it's practival.
Given you can only return one type from each branch, you are likely to either need span in all or none of them.

zesterer:
I think having an API that looks visually consistent and obvious is more important than needing to type a few extra characters, personally. Perhaps there is an alternative design that fulfils both objectives, but I don't really see it.

JohnDowson:
Perhaps something like this?
```rust
select! {
    span => ident,
    Token::Word(w) => AstNode { span, ast: AstKind::Word(w) }
    Token::KeyWord(kw) => AstNode { span, ast: AstKind::KeyWord(kw) }
}
```

zesterer:
That's ambiguous syntax (`span` is itself a valid pattern).

JohnDowson:
> That's ambiguous syntax (`span` is itself a valid pattern).

Well, you could use `;` instead of `,` for span binding. Either way, that's your library, I'll be happy if this feature exists in whatever form you choose.

zesterer:
I've implemented this in 0d7e72a along with an example. Hopefully this is sufficient for your use-case!

PR #104, opened 2022-03-13
[zero-copy] Re-add `no_std` support, add `separated_by_exactly`, and a few primitives

wackbyte:
Adds the `any`, `empty`, `one_of` primitives, the `separated_by_exactly` combinator, and `no_std` support back.

### DIFF
--- a/benches/json.rs
+++ b/benches/json.rs
@@ -122,10 +122,7 @@ mod chumsky_zero_copy {
 
             let array = value
                 .clone()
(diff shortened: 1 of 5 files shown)


zesterer:
Thanks!

wackbyte:
for posterity, here are the benchmarks after using the new `separated_by`:
image
compared to previously, where a combo of other combinators was used in place of it:
image

PR #106, opened 2022-03-15
[zero-copy] Add `rewind` and `then_with` combinators

wackbyte:


zesterer:
Thanks! Reminds me, I really need to start working on error propagation.

Issue #77, opened 2022-01-24
Feature request: Allow users to implement custom recovery strategies


jfecher:
As it stands now, users only have access to `skip_until`, `skip_then_retry_until`, and `nested_delimiters`. I'm unsure if the best way to go about this would be exposing the `Strategy` trait (which currently uses private types) - preferably in a cleaner form, or adding lower-level recovery strategies/parsers that can be used to build up more complex strategies. As it stands, I believe it isn't possible to implement a strategy like `expect` from https://eyalkalderon.com/blog/nom-error-recovery/. `skip_until` is close but provides no way to not skip any tokens.

jfecher:
I've been thinking a bit on the correspondence between recovery strategies and parsers themselves. For starters, recovery strategies appear to be normal parsers that are called on otherwise failing input. In that respect it'd be a shame if we couldn't reuse normal parsers with them and build up bigger recovery strategies the same way bigger parsers are built up. Using this, I think the simplest error recovery strategy may be a simple `a.or(b)`-like strategy in which if `a` fails, the error is still logged but the parser `b` is tried in an attempt to recover from any errors.

Using this function we can build up more complicated error recovery strategies like the `expect` mentioned above. Since an `or` parser already exists, I'll use `recover_with` instead, but taking in any Parser, rather than any Strategy. If that second parser successfully parses, the parse is recovered:
```rs
let nothing = just([]).or_not();
let expect = |p, default| p.recover_with(nothing, default);

just(Let)
    .then(expect(ident(), Ident::error))
    .then(expect(just(Equals), ...)) //etc.
    
// Inverts a parser to succeed when it would otherwise fail, and vice-versa
let not = |parser| parser.try_map(|value, span| {
    match value {
        Ok(_) => Err(...),
        Err(_) => Ok(()),
    }
});

// skip_until is easily definable
let skip_until = |tokens, f| not(just(tokens)).repeated().map_with_span(|_, span| f(span));
```

zesterer:
Apologies about not replying to this earlier, GitHub decided to omit a swathe of things from my notifications for some reason.

I have been having similar thoughts about this too. I'm very tempted to take another swing at error recovery, with a more parser-driven approach. If I get time over the next week, I'll do some experimenting and see what I can come up with!

Alternatively, if you have any specific thoughts (I like the `expect` notation) I'd love to hear them!

jfecher:
[...]

PR #107, opened 2022-03-15
Add 'recover_via' parser

jfecher:
Adds a new parser (currently named `recover_via` - I'm open to suggestions) which attempts the first parser and returns its result on success. On error it will try the given recovery parser and if that succeeds it will return that result but with the errors from the first parser. In the case both parsers fail, the second is considered to be a failed recovery strategy and only the errors from the first are returned. This allows you to chain multiple successive `recover_via` calls.

This parser is similar to `Parser::or` except that it returns the errors of the first parser even if the second succeeds.

This parser is also similar to `recover_with(Strategy)` though it accepts any parser as an argument rather than only an existing strategy, and is thus more general and easier to extend for users (at least with the current design of Strategy).

### DIFF
--- a/src/recovery.rs
+++ b/src/recovery.rs
@@ -361,9 +361,10 @@ impl<I: Clone, O, A: Parser<I, O, Error = E>, B: Parser<I, O, Error = E>, E: Err
         #[allow(deprecated)]
         let (mut a_errors, a_out) = debugger.invoke(&self.0, stream);
 
-        if a_out.is_ok() {
-            return (a_errors, a_out);
-        }
+        let a_error = match a_out {
+            Ok(_) => return (a_errors, a_out),
+            Err(error) => error,
+        };
 
         stream.revert(pre_state);
 
@@ -371,12 +372,13 @@ impl<I: Clone, O, A: Parser<I, O, Error = E>, B: Parser<I, O, Error = E>, E: Err
         let (mut b_errors, b_out) = debugger.invoke(&self.1, stream);
 
         if b_out.is_ok() {
+            a_errors.push(a_error);
             a_errors.append(&mut b_errors);
             return (a_errors, b_out);
         }
 
         stream.revert(pre_state);
-        (a_errors, a_out)
+        (a_errors, Err(a_error))
     }
 
     fn parse_inner_verbose(&self, d: &mut Verbose, s: &mut StreamOf<I, E>) -> PResult<I, O, E> {


src/recovery.rs
```diff
@@ -341,3 +341,55 @@ impl<I: Clone, O, A: Parser<I, O, Error = E>, S: Strategy<I, O, E>, E: Error<I>>
...
+        stream: &mut StreamOf<I, E>,
+    ) -> PResult<I, O, E> {
+        let pre_state = stream.save();
+
+        #[allow(deprecated)]
+        let (a_errors, a_out) = debugger.invoke(&self.0, stream);
+
+        if a_errors.is_empty() {
+            if let Ok(a_out) = a_out {
+                return (vec![], Ok(a_out));
```
zesterer, line 366: Perhaps return `a_errors` here to take advantage of any existing allocated vector space?

zesterer:
Thanks for the PR! I like this change overall, and I was planning for something similar to become the basis for error recovery in the `zero-copy` branch. I wonder whether this would be best implemented as a recovery strategy though: something like `.recover_with(parser(...))`? Interested to hear your thoughts.

src/recovery.rs
jfecher: Good idea, fixed.

jfecher:
I was actually thinking the opposite, that we could get rid of `Strategy` altogether in favor of only having `recover_via` or other parsers implemented in terms of it. `skip_until` for example could be redefined as `recover_via(none_of(just(xs)).repeated())`, though it'd still need an end of input check to fail. Using this approach over strategy simplifies the API for users a bit in my eyes, and lets the future of the crate focus on adding more parsers rather than splitting efforts between parsers and recovery strategies.

zesterer:
Yep, this was my plan for `zero-copy`: `nested_delimiters` would then just become a parser primitive. However, I'd rather avoid splitting work between `master` and `zero-copy` for now (it's almost a complete refactor of the crate). Perhaps I can merge this MR, but it might not end up in a stable release, at least before `zero-copy` is merged. Do you have thoughts on that?

jfecher:
This PR implements #77 where the `recover_with` parser given there is defined here as `recover_via`

zesterer:
Okay. Thanks for the PR though, it's useful to be able to point people at it!

src/recovery.rs
```diff
@@ -341,3 +341,50 @@ impl<I: Clone, O, A: Parser<I, O, Error = E>, S: Strategy<I, O, E>, E: Error<I>>
...
+            return (a_errors, a_out);
+        }
+
+        stream.revert(pre_state);
+
+        #[allow(deprecated)]
+        let (mut b_errors, b_out) = debugger.invoke(&self.1, stream);
+
+        if b_out.is_ok() {
+            a_errors.append(&mut b_errors);
```
zesterer, line 374: I guess the question of whether secondary errors from b should be preserved is an open one, but I'd err on the side of saying that they shouldn't. Not certain though.

zesterer:
This is effectively superseded by #233 now. I'm happy to reopen this if things change, however.

PR #110, opened 2022-03-18
[zero-copy] Implement missing primitives

CraftSpider:
May need more testing, but these should be all the primitives in main but not zero-copy except the deprecated `seq`

### DIFF
--- a/src/zero_copy/primitive.rs
+++ b/src/zero_copy/primitive.rs
@@ -423,10 +423,10 @@ where
 
         loop {
             let start = inp.save();
-            match self.until.go::<M>(inp) {
-                Ok(out) => break Ok(M::combine(output, out, |output, out| (output, out))),
-                Err(_) => ()
+            if let Ok(out) = self.until.go::<M>(inp) {
+                break Ok(M::combine(output, out, |output, out| (output, out)));
             }
+
             inp.rewind(start);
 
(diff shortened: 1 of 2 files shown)


src/zero_copy/primitive.rs
```diff
@@ -318,6 +318,159 @@ pub fn any<I: Input + ?Sized, E: Error<I::Token>>() -> Any<I, E> {
...
+
+    fn go<M: Mode>(&self, inp: &mut InputRef<'a, '_, I, E, S>) -> PResult<M, Self::Output, E> {
+        let mut output = M::bind(|| C::default());
+
+        loop {
+            let start = inp.save();
+            match self.until.go::<M>(inp) {
+                Ok(out) => break Ok(M::combine(output, out, |output, out| (output, out))),
+                Err(_) => ()
+            }
```
zesterer, line 429: `if let Ok(out) = ...`?

CraftSpider: You've caught my most frequent mistake - When I refactor a match and end up with just an if-let, but never change it over. Fixing now

wackbyte:
Can you add them to the prelude as well?

wackbyte (approved):

zesterer:
Thanks for the PR!

PR #123, opened 2022-04-15
Make defining recursive twice error

CraftSpider:


bew:
Why do you want to disallow this?

CraftSpider:
The existing panic in the recursive `define` method implies it is meant to be an error to define it twice. Doing so overwrites the first usage and is, in most situations, I imagine an undesired behavior (And, if OnceCell is stabilized in the future, possibly becomes impossible at that point anyways). If there's a use case, the panic message could instead be changed, and the implementation could simply use a RefCell instead of a OnceCell.

zesterer:
Thanks!

PR #124, opened 2022-04-16
fix tutorial code

taka231:
I found that the code in the tutorial (Parsing parentheses) has an error.
That code doesn't work in the case: "3 * (4 + 2)".
By adding .padded(), that code works well.

### DIFF
--- a/tutorial.md
+++ b/tutorial.md
@@ -404,7 +404,7 @@ recursive(|expr| {
         .padded();
 
     let atom = int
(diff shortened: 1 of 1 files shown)


zesterer:
Thanks, although I think this `.padded()` should go at the end of `atom` instead so that it works recursively.

taka231:
OK, I added .padded() at the end of atom.

zesterer:
Thanks!

Issue #127, opened 2022-04-18
Reveal the fact that `whitespace` is a `Repeated` to allow calling methods like `at_least` on it


zesterer:

zesterer:
Good point!

PR #129, opened 2022-04-19
issue #127 - repeated whitespace()

dis-da-moe:
My attempt at [this issue](https://github.com/zesterer/chumsky/issues/127). I tried to keep the `custom` parser, but I wasn't sure how that could be implemented as a `Repeated`, so I used `filter` instead. This might be breaking because the return is now a `Vec<()>` instead of just `()`, but it can be used with an `ignored()` at the end to be like the original.

```rust
pub fn whitespace<'a, C: Character + 'a, E: Error<C> + 'a>(
) -> Repeated<impl Parser<C, (), Error=E>+Copy+Clone + 'a>{
    filter(|c: &C| c.is_whitespace()).ignored().repeated()
}
```

### DIFF
--- a/src/text.rs
+++ b/src/text.rs
@@ -177,7 +177,9 @@ impl<I: Character, O, P: Parser<I, O>> TextParser<I, O> for P {}
 
 /// A parser that accepts (and ignores) any number of whitespace characters.
 ///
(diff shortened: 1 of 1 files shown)


zesterer:
This looks about right, thanks!

PR #134, opened 2022-05-06
Fix wording of documentation

MiSawa:
Thank you for a grate crate!

This PR replaces 2 occurrences of "positive integer" in the documentation with "non-negative integer", because `text::int` actually do parse `"0"` correctly as the doc says.


zesterer:
Thanks!

PR #137, opened 2022-05-10
Implement Eq for Simple error

remimimimimi:
Need this to use chumsky with [salsa](https://github.com/salsa-rs/salsa).


zesterer:
Thanks!

PR #140, opened 2022-05-15
[zero-copy] Add state combinators

CraftSpider:
Adds two simple combinators for accessing the parse state

### DIFF
--- a/src/zero_copy/primitive.rs
+++ b/src/zero_copy/primitive.rs
@@ -423,7 +423,7 @@ pub fn filter_map_state<F, I, O, E, S>(filter: F) -> FilterMapState<F, E>
     where
         I: Input + ?Sized,
         E: Error<I>,
-        F: Fn(&mut S, &I::Token) -> Result<O, E>,
+        F: Fn(&mut S, I::Span, &I::Token) -> Result<O, E>,
 {
     FilterMapState {
         filter,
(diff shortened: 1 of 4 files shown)


src/zero_copy/primitive.rs
```diff
@@ -404,6 +404,63 @@ where
     go_extra!();
 }
 
+pub struct FilterMapState<F, E> {
```
zesterer, line 407: I agree that this is a useful combinator, but the name conflict with the existing `filter` and `filter_map` primitives make me uncomfortable. That said, I think that it's those primitives that were named in error, and not this combinator. What do you think about removing those primitives entirely and replacing them with `any().filter(..)`/`any().filter_map(...)`?

CraftSpider: That sounds good to me. It seems kind of odd that you 'filter' as a first step. I feel like I've definitely reached for filter as a trait method and had to remind myself it wasn't at least once or twice.

CraftSpider: With that, `try_map_state` might be a better name, at least for my use case

zesterer: Perhaps `try_map_with_state`, to mirror `map_with_span`?

CraftSpider: That works. Also feel like this is hitting the std situation: params x, y, and z, do we add all the superset of combinations, or how do we pick which should be included :P

zesterer:
Thanks!

PR #148, opened 2022-05-28
Don't skip token in SkipThenRetryUntil

Jim-Eckerlein:
Previously, `SkipThenRetryUntil::recover()` would skip a token if it is not in the array used as stopper-tokens (`SkipThenRetryUntil::0`). That means that the retry-parser would only start at the next token. That behaviour is incorrect, as the skipped token might already be the one which should be parsed.

I added unit tests which fail without 1a0b8cff4efb05a21194295d1568759eacabd896


zesterer:
Thanks for the PR!

I agree that this behaviour is unexpected. I'm fairly sure I recall adding this line for a reason though, but I forgot what it was (should have added a comment!).

I'll merge this for now, but it might be that I end up reverting it if users report breakage. Regardless, I'm not particularly happy with the crate's current error recovery strategy, so it's probably going to get an overhaul in the `zero-copy` branch.

PR #150, opened 2022-06-03
Support arrays for choice

CraftSpider:
This allows usage of arrays of any length in choice - inspired by a need in a personal project where I have more than 26 choices, which can all be represented as one type of parser easily. It may be worth it to note that adding `boxed` will allow any set of parsers to share a type.

### DIFF
--- a/src/primitive.rs
+++ b/src/primitive.rs
@@ -887,6 +887,51 @@ impl<T: Clone, E> Clone for Choice<T, E> {
     }
 }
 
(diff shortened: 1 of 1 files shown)


zesterer:
Hmm, interesting! Perhaps this could be added for `Vec` too, more easily allowing the dynamic construction of parsers?

zesterer (approved):

CraftSpider:
Hm, I was going to say implement it on slices, but I realized that would quickly lead to frustration with lifetimes. So if you want to support dynamic construction, `Vec` is probably a good idea. I personally haven't done much dynamic construction of parsers, but I kind of like that idea, I can see a use case for dynamic generation of choices, for example in a language with dynamic syntax. I'll add an impl.

PR #151, opened 2022-06-04
Fixes #131 using Debug format to display errors

### LINKED ISSUE #131: `Simple` display should show `Debug` representations of input / expectations.
The `Display` impl for `Simple` shows raw output and expectations, which is awkward for whitespace, such as '\n':

Example:

```
use chumsky::primitive::{end, just};
use chumsky::Parser;

type Error = chumsky::error::Simple<char>;

fn main() {
    match parser().then_ignore(end()).parse("\nblah") {
        Ok(_) => panic!("Should not happen."),
        Err(es) => {
            for e in es {
                println!("{}", e);
            }
        }
    }
}

fn parser() -> impl Parser<char, (), Error = Error> {
    just("foo")
        .ignored()
        .or(just("\r\n").repeated().ignored())
        .or(just(' ').repeated().at_least(1).ignored())
}
```

-gives output:

```
found '
 was expectedend of input, f, 
```

Note also that sending raw whitespace (either inputs or expectations) behaves differently on different platforms and terminal types.

A `Debug`-style representation would be more useful for users, something like:

```
found '\n', was expecting 'f', '\r', or '\n'
```

dark-fusion:
This change introduces a Debug-style representation for `Simple`. It
should close issue #131 as the output now closely matches the desired
output discussed in the issue.

There may be some edge-cases or areas for improvement. In that case,
please let me know and I'll be happy to revise the PR.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -309,32 +309,31 @@ impl<I: fmt::Display + Hash + Eq, S: Span> fmt::Display for Simple<I, S> {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         // TODO: Take `self.reason` into account
 
-        let found = if let Some(found) = &self.found {
-            format!("{}", found)
+        if let Some(found) = &self.found {
+            write!(f, "found {:?}", found.to_string())?;
         } else {
-            format!("end of input")
+            write!(f, "found end of input")?;
         };
 
         match self.expected.len() {
             0 => {} //write!(f, " but end of input was expected")?,
             1 => write!(
                 f,
-                " but {} was expected",
+                " but expected {}",
                 match self.expected.iter().next().unwrap() {
-                    Some(x) => format!("'{}'", x),
-                    None => format!("end of input"),
+                    Some(x) => format!("{:?}", x.to_string()),
+                    None => "end of input".to_string(),
                 },
             )?,
             _ => {
                 write!(
                     f,
-                    "found {:?}, but one of {:?} was expected",
-                    found,
+                    " but expected one of {}",
                     self.expected
                         .iter()
                         .map(|expected| match expected {
-                            Some(x) => format!("{}", x),
-                            None => found.to_string(),
+                            Some(x) => format!("{:?}", x.to_string()),
+                            None => "end of input".to_string(),
                         })
                         .collect::<Vec<_>>()
                         .join(", ")
--- a/src/error.rs
+++ b/src/error.rs
@@ -312,7 +312,7 @@ impl<I: fmt::Display + Hash + Eq, S: Span> fmt::Display for Simple<I, S> {
         let found = if let Some(found) = &self.found {
             format!("{}", found)
         } else {
-            format!("found end of input")
+            format!("end of input")
         };
 
         match self.expected.len() {
@@ -328,7 +328,7 @@ impl<I: fmt::Display + Hash + Eq, S: Span> fmt::Display for Simple<I, S> {
             _ => {
                 write!(
                     f,
-                    r"found {:?}, but one of {:?} was expected",
+                    "found {:?}, but one of {:?} was expected",
                     found,
                     self.expected
                         .iter()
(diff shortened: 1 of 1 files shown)


src/error.rs
```diff
@@ -309,11 +309,11 @@ impl<I: fmt::Display + Hash + Eq, S: Span> fmt::Display for Simple<I, S> {
...
         // TODO: Take `self.reason` into account
 
-        if let Some(found) = &self.found {
-            write!(f, "found '{}'", found)?;
+        let found = if let Some(found) = &self.found {
+            format!("{}", found)
         } else {
-            write!(f, "found end of input")?;
-        }
+            format!("found end of input")
```
bew, line 315:
```suggestion
            format!("end of input")
```
otherwise you'd get `found found end of input ...`

src/error.rs
```diff
@@ -325,18 +325,21 @@ impl<I: fmt::Display + Hash + Eq, S: Span> fmt::Display for Simple<I, S> {
...
-            )?,
+            _ => {
+                write!(
+                    f,
+                    r"found {:?}, but one of {:?} was expected",
+                    found,
+                    self.expected
+                        .iter()
+                        .map(|expected| match expected {
+                            Some(x) => format!("{}", x),
```
bew, line 336: don't you need `{:?}` here ?

src/error.rs
dark-fusion: Definitely true. I agree with this change. I overlooked that part.

src/error.rs
dark-fusion: This is the tricky bit. I played around with it for a bit and you can use `{:?}` here, but you'll get very strange output. The escape `\` tokens are included in the format string.

There seems to be several different ways one could do this. I want to introduce a solution that doesn't involve messing with trait bounds or anything else. I'm sure there is a more elegant and/or more optimized way to handle this.

bew: I mean, did you try to put `{:?}` here and remove leave a simple `{}` line 331 for the `.. one of {} was expected`

otherwise I guess it'll use the debug output of the debug output, leading to many `\`

src/error.rs
```diff
@@ -325,18 +325,21 @@ impl<I: fmt::Display + Hash + Eq, S: Span> fmt::Display for Simple<I, S> {
...
+            _ => {
+                write!(
+                    f,
+                    "found {:?}, but one of {:?} was expected",
+                    found,
+                    self.expected
+                        .iter()
+                        .map(|expected| match expected {
+                            Some(x) => format!("{}", x),
+                            None => found.to_string(),
```
bew, line 337: I'm wondering, why use `found` here?

src/error.rs
```diff
@@ -325,18 +325,21 @@ impl<I: fmt::Display + Hash + Eq, S: Span> fmt::Display for Simple<I, S> {
                     None => format!("end of input"),
```
bew, line 325: and I think you missed a change in the few lines above this line (but I can't put a comment there)

src/error.rs
dark-fusion: I'll take a closer, more careful look and implement things as per your suggestions. I may have missed something crucial because my intuition is telling me that your approach is correct.

src/error.rs
dark-fusion: I'm not very familiar with the codebase and I suppose I tried to make as small a change as possible. I don't like this, either. I don't have a good explanation for this.

With that said, I will clean this section up. Thank you for pointing this out.

src/error.rs
dark-fusion: Got it.

I really appreciate your time and your help. Thank you for reviewing this with me. I will make the changes that you suggest.

zesterer:
Seems like @bew has reviewed this already! Many thanks to you both. Ping me when you feel it's ready!

dark-fusion:
@zesterer The PR should be good to go. I made the suggested changes and pushed them. Please feel free to let me know if anything doesn't look right.

Thanks!

zesterer:
Thanks for the PR! I might make an attempt to reduce instances of allocation later, but for now I'm happy to accept it.

nathan-at-least:
Thanks for these two fixes, which address the main `chumsky` problems I have in my current project. When will this PR be included in a new release?

zesterer:
@nathan-at-least I need to find the time/energy to prepare a new release. Perhaps this week some time. For now, you can point your `Cargo.toml` to the git repository.

PR #152, opened 2022-06-06
[zero-copy] make map_with_state consitent with other maps / try_map_with_state

CraftSpider:
Make map_with_state non-fallible, and elide it in check mode like `map` or `map_with_span`

This may be more iffy, as the user may mutate the state. One option would be to make `map_with_state` pass in a `&S` instead of `&mut S`, to indicate that it is supposed to read-only the state, and force `try_map_with_state` if you rely on the output.

### DIFF
--- a/src/zero_copy/combinator.rs
+++ b/src/zero_copy/combinator.rs
@@ -156,8 +156,8 @@ impl<'a, I, E, S, A, F, O> Parser<'a, I, E, S> for MapWithState<A, F>
 
     fn go<M: Mode>(&self, inp: &mut InputRef<'a, '_, I, E, S>) -> PResult<M, Self::Output, E> {
         let before = inp.save();
-        self.parser.go::<M>(inp).map(|out| {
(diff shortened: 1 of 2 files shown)


zesterer:
Hmm, I'm not convinced that this is a good idea, at least by default. Having this fail could potentially result in very surprising behaviour, and it would be difficult for users to determine why it happened, or even search for the problem at all. I would rather that it worked by default. Do you need this for something in particular?

CraftSpider:
No, I was just considering which places explicitly called `Emit`, and I noticed this idiosyncracy between this and other maps. Maybe it would be better to add some `map_with_state_pure` which behaves like this? Not including it at all is fine, I was just looking for a chance to improve efficiency. Even without the emission change, I think the signature change on its own is good, and that lack of `&mut` may actually be a bug.

CraftSpider:
Updated to only include the signature change

zesterer:
Thanks!

PR #154, opened 2022-06-07
Add OrderedContainer, update some trait docs

CraftSpider:
Now `just` cannot be used with HashSets or Heaps (while technically ordered, a Heap is unlikely to do what the user wants)


zesterer:
Thanks!

Issue #157, opened 2022-06-13
[Question] Printing with ariadne


TornaxO7:
Hi! I'm trying to use [`ariadne`](https://docs.rs/ariadne/0.1.5/ariadne/index.html) to get a nice error message from `chumsky` but I'm struggeling with the message.

How am I supposed to use [`SimpleReason`](https://docs.rs/chumsky/0.8.0/chumsky/error/enum.SimpleReason.html)? I expected to be able to do it like this:
```rust
            Report::build(ReportKind::Error, Self::ERROR_ID, Self::ERROR_OFFSET)
                .with_message(error.reason())
```
but `error.reason()` returns me a `SimpleReason` which isn't printable? How am I supposed to use `error.reason()`?

TornaxO7:
> You can match on SimpleReason and decide how to display it yourself,

Oh ok.

> although I probably should add a default Display implementation for it. 

that would be nice to have :)

PR #158, opened 2022-06-13
adding default display implementation for SimpleReason

TornaxO7:
A possible default `Display` implementation for `SimpleReason`(?)

Comes from https://github.com/zesterer/chumsky/issues/157

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -166,14 +166,12 @@ pub enum SimpleReason<I, S> {
     Custom(String),
 }
 
-impl<I, S> SimpleReason<I, S> {
-    const DEFAULT_DISPLAY_UNEXPECTED: &'static str = "unexpected value";
-}
-
 impl<I: fmt::Display, S: fmt::Display> fmt::Display for SimpleReason<I, S> {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        const DEFAULT_DISPLAY_UNEXPECTED: &'static str = "unexpected input";
+
         match self {
-            Self::Unexpected => write!(f, "{}", Self::DEFAULT_DISPLAY_UNEXPECTED),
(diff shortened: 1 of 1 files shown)


src/error.rs
```diff
@@ -166,6 +166,20 @@ pub enum SimpleReason<I, S> {
     Custom(String),
 }
 
+impl<I, S> SimpleReason<I, S> {
+    const DEFAULT_DISPLAY_UNEXPECTED: &'static str = "unexpected value";
```
zesterer, line 170: I think it's probably find to just have this inline within the implementation of `Display`. I would probably change it to `"unexpected input"` though, in order to retain consistency with other parts of the codebase.

zesterer:
Thanks! Just one comment.

src/error.rs
TornaxO7: > I think it's probably find to just have this inline within the implementation of Display.

I'd prefer to have this as a constant here in order to avoid [magic strings](https://en.wikipedia.org/wiki/Magic_string). But if you wish I can put it inline into the implementation of the `Display` trait.

> I would probably change it to "unexpected input" though, in order to retain consistency with other parts of the codebase.

Understandable, I'll change that ~~tomorrow~~ today, if I've slept.

zesterer: > I'd prefer to have this as a constant here in order to avoid [magic strings](https://en.wikipedia.org/wiki/Magic_string)

I'm not sure this is the same as a magic string. The text doesn't have any particularly special meaning outside of the `Display` impl itself, does not result in unusual behaviour, and the constant is neither configurable not publicly visible anyway. There's also another similar string already in the `Display` impl already.

TornaxO7: done

zesterer:
Thanks!

Issue #20, opened 2021-11-03
Example of Python-like parsing


VinnyVicious:
Having an example on how to parse Python-like languages that are aware of indentation would be interesting.

zesterer:
Python's parser/lexer (whichever processes identations) is not context-free because correctly parsing each line relies on the contextual information available in previous lines (i.e: lines 'pair up' with neighbouring lines into blocks if their indentation matches).

Chumsky (and virtually all parser combinators and parser generators) do not play well with context-sensitive grammars. It's usually necessary to hand-write a context-sensitive parser for such languages.

All is not lost, however! A neat trick you can use is to write your lexer by hand and have it translate Pythonic indentation into traditional code blocks with delimiters. Once this is done, the tokens can be fed into Chumsky and parsed in a context-free manner.

I would actually advise you use this strategy: get rid of the context-sensitive indentation as early as you possibly can in the compiler, it makes parsing in general needlessly complex.

I've actually written an example lexer that does this in the past, complete with correct handling of parentheses and newlines. You can take a look at it [here](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=39f089d64a36f8535c4696170e8b6e0e). Once you've turned the text into a stream of tokens, they're pretty easy to feed to Chumsky.

It's probably possible to implement something similar as custom parser for Chumsky too, although it's obviously a bit ugly for the reasons I mentioned above. I'll look into this a little more to see if it's possible.

VinnyVicious:
Thank you for the detailed answer and the bonus example! Really appreciate it.

Xandaros:
Something I have seen before is introducing special `INDENT` and `OUTDENT` tokens in your lexer. I thought that was quite clever.

LikeLakers2:
@zesterer Apologies for the bump after two weeks, but I'm curious.

> I'll look into this a little more to see if it's possible.

If it turns out not to be possible, could an indent-aware parsing example still be included in the `/examples/` directory? That way, users would know how they might make Chumsky work with indent-aware content, even if it requires some hand-written code on their end.

zesterer:
I've had a think about it and although I believe that it should be possible to implement, I don't think that any implementation I could add to Chumsky would be sufficiently flexible to cover all use-cases.

In terms of a hand-written example implementation, I'm very happy to accept something like this if someone is interested in implementing it. Right now, unfortunately, I'm focusing on a number of other things so I likely won't get time to work on it until after New Year.

[...]

PR #166, opened 2022-07-06
Update the pythonic example to demonstrate how to create a flatten token stream from a token tree

tatsuya6502:
This PR adds a step to the pythonic example to convert a token tree from the lexer into a flat token stream using `Stream::from_nested`.

This would not be a long-term solution because it's quite verbose, but it will work fine for now.

Relates to #20.

### DIFF
--- a/examples/pythonic.rs
+++ b/examples/pythonic.rs
@@ -1,4 +1,17 @@
-use chumsky::prelude::*;
+use chumsky::{prelude::*, BoxStream, Flat, Stream};
+
+type Span = std::ops::Range<usize>;
+
+#[derive(Clone, Debug)]
+enum Token {
+    Int(u64),
+    Ident(String),
+    Op(String),
+    OpenParen,
+    CloseParen,
+    BeginBlock,
+    EndBlock,
+}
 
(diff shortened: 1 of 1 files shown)


zesterer:
Thanks! I wonder whether the utility functions could be placed inline at their use site give that they're not particularly complex? I think it makes it easier to see what's going on (i.e: avoids confusion about them being functions from chumsky)

tatsuya6502:
Thank you for reviewing the code! I will update my code in this weekend.

zesterer:
@tatsuya6502 Are you happy for me to take this branch and make the changes myself?

zesterer:
I've used the work in this PR as a reference when improving the example in 91e5cc8d7912a77233144aee8209891dc98cb3af. Thanks!

tatsuya6502:
@zesterer Thank you for updating the code and merging it! It is silly; I have been distrusted by other stuff recently. So I am very happy to see the code was merged. Thank you!

zesterer:
Thanks so much for the initial work you did, it was very useful as a reference!

PR #183, opened 2022-08-10
fix: change core::lazy::OnceCell to core::cell::OnceCell

dawnofmidnight:
I don't think this needs any further explanation.


zesterer:
Thanks!

PR #190, opened 2022-08-13
Zero copy cleanup

Zij-IT:
Decided since I am super excited for the Zero-Copy implementation that is up-and-coming, I would try an contribute a little back in the form of some small cleanup commits.

Most of these were just some bits that I found while perusing through the code, and thought they could be done a tad simpler. Let me know if I missed something, or if something needs changing!

Thanks for your time!

### DIFF
--- a/src/zero_copy/input.rs
+++ b/src/zero_copy/input.rs
@@ -181,12 +181,9 @@ impl<'a, 'parse, I: Input + ?Sized, E: Error<I>, S> InputRef<'a, 'parse, I, E, S
     pub(crate) fn skip_while<F: FnMut(&I::Token) -> bool>(&mut self, mut f: F) {
         loop {
             let before = self.save();
-            match self.next() {
-                (_, Some(c)) if f(&c) => {}
-                (_, Some(_) | None) => {
-                    self.rewind(before);
-                    break;
-                }
+            if self.next().1.filter(&mut f).is_none() {
+                self.rewind(before);
+                break;
             }
         }
     }
--- a/src/zero_copy/input.rs
+++ b/src/zero_copy/input.rs
@@ -35,13 +35,8 @@ impl Input for str {
     }
 
     fn next(&self, offset: Self::Offset) -> (Self::Offset, Option<Self::Token>) {
-        if offset < self.len() {
-            // TODO: Can we `unwrap_unchecked` here?
-            let c = unsafe { self.get_unchecked(offset..).chars().next().unwrap() };
-            (offset + c.len_utf8(), Some(c))
-        } else {
-            (offset, None)
-        }
+        let chr = self.chars().skip(offset).next();
+        (offset + chr.map_or(0, char::len_utf8), chr)
     }
 
     fn span(&self, range: Range<Self::Offset>) -> Self::Span {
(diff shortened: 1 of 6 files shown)


zesterer:
Hey, this looks like great work! Is there any chance you could switch the target branch to `zero-copy`?

Zij-IT:
Whoops! Totally thought I had it set to zero-copy. Its fixed!

src/zero_copy/input.rs
```diff
@@ -35,13 +35,8 @@ impl Input for str {
...
     fn next(&self, offset: Self::Offset) -> (Self::Offset, Option<Self::Token>) {
-        if offset < self.len() {
-            // TODO: Can we `unwrap_unchecked` here?
-            let c = unsafe { self.get_unchecked(offset..).chars().next().unwrap() };
-            (offset + c.len_utf8(), Some(c))
-        } else {
-            (offset, None)
-        }
+        let chr = self.chars().skip(offset).next();
+        (offset + chr.map_or(0, char::len_utf8), chr)
```
zesterer, line 39: I'd prefer if this change were not made because this is extremely hot code. Even single branches will have a substantial impact on parser performance (ideally. `unwrap_unchecked` would be used too: `str` is guaranteed to always be valid UTF-8, so provided we can guarantee that `offset` is valid, this should be fine).

zesterer:
Thanks for the PR! I'm broadly in favour of this, except the change I commented on. Thanks!

src/zero_copy/input.rs
Zij-IT: With your comment in mind, I went ahead and ran both variations through Godbolt , and came to the conclusion that my suggestion had about 3x the amount of branches that yours did :D So, good call there. I was curious though what a hybrid variation would look like, and ended up doing this:

```rust
fn next(&self, offset: Self::Offset) -> (Self::Offset, Option<Self::Token>) {
    let chr = unsafe { self.get_unchecked(offset..).chars().next() };
    (offset + chr.map_or(0, char::len_utf8), chr)
}
```

The output of Godbolt can be found [here](https://rust.godbolt.org/z/3oxh9YhTs), and shows that this variation has one less branch than the original implementation. I am curious on if you would have any benchmarks that could be used to determine which of the two implementations proves faster!

zesterer: That seems like it might be better in that case! The `json` benchmark is currently the canonical way to benchmark chumsky (although it would be nice to have others in the future given that it doesn't include cases like backtracking).

Zij-IT:
I have added the hybrid variant in [this commit](https://github.com/zesterer/chumsky/pull/190/commits/2f7074238c35ef7beb39fb70bcb147d1adb848ca)! Let me know if any more changes are necessary!

zesterer:
Thanks very much, I really appreciate you taking the time to work on chumsky! Hopefully this pushes `zero-copy` incrementally closer to being ready to replace `master` :)

Zij-IT:
Happy to help! It has helped me to better understand how such parsers work, and was amazing to work with. If you need more helping hands in the future, don't be afraid to tag me :D

src/zero_copy/input.rs
Zij-IT: I just realized why it would have one less branch, and I think we both missed something when we thought we could merge it. I believe we have to revert the changes here.

Assuming the called passes in an offset that is too large, its no longer going to return none, but instead read memory that it shouldn't. I spaced that `get_unchecked` couldn't be used without a check, so that's my bad.

PR #198, opened 2022-08-27
Random `zero-copy` fixes!

Zij-IT:
While I was moving my lexer over to `zero-copy` there were some issues that I encountered related to `TakeUntil`, and I am sad to say it took like an hour and a half of dealing with Rusts wonderful generic errors before I was able to find the reason why.

The current implementation for of `Parser` is only implemented  for `TakeUntil<P, C>`, and when one fills in the missing items, you get: `TakeUntil<P, C, (), (), ()>`. This was a particularly fun problem because `C`, which represents the collection that should be parsed into, is actually in the spot for the input generic `I`, which we can see when we look at the definition here `pub struct TakeUntil<P, I: ?Sized, C = (), E = (), S = ()>`

Additionally I fixed the bug that I accidentally introduced in one of my previous pull requests. At the time I believed it to be correct, and have gotten one less branch. What I didn't realize was that I removed the branch that assured that the `offset` index is within the `&str`. By removing that I enabled the classic "read memory that doesn't belong to you" bug.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,7 +2,7 @@
 name = "chumsky"
 version = "0.8.0"
 description = "A parser library for humans with powerful error recovery"
-authors = ["Joshua Barretto <joshua.s.barretto@gmail.com>"]
+authors = ["Joshua Barretto <joshua.s.barretto@gmail.com>", "Elijah Hartvigsen <elijah.reed@hartvigsen.xyz"]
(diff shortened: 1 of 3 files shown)


zesterer:
Thanks! It's just occurred to me that there should probably be safe and unsafe variants of `Input::next`, given that it's a public trait.

PR #203, opened 2022-09-23
zero-copy: several things

### LINKED ISSUE #194: request: make most/all of the primitive constructor functions into const fn
Allow 
```rust
const OPEN_PAREN: Just<char, char, Error=Simple<char>> =
  just('(');
```
and such

### LINKED ISSUE #199: n-tuples as output of multiple .then() parsers
If I have a chain of `then` parsers, e.g.:

```

parse_int()
.then_ignore(parse_whitespace())
.then(parse_char())
.then_ignore(parse_whitespace()
.then(parse_float().then(parse_float()))
```

The output type of this parser will be a nested sequence of 2-tuples, i.e. `((i32, char), (f32, f32))`.

When using `.map` on the output of this kind of built-up parser, it's very easy to get confused about the exact internal structure of the 2-tuples, especially when there are multiple levels of nested parser.

It would be convenient if chumsky had a combinator that could accept multiple parsers in sequence of heterogenous output type, and produce an arbitrarily-sized tuple as output. This is similar to the behavior of the [tuple](https://docs.rs/nom/latest/nom/sequence/fn.tuple.html) combinator in nom.

wackbyte:
Changes:
- Remove the `generic_associated_types` feature as it's stable on nightly
- `const`-ify primitive creation functions
  - Closes #194
- Add the `Group` primitive (name ideas welcome!)
  - Minimally tested, may need work
  - Not yet sure how runtime/compiletime performance compares to `then` chains
  - `group` is to `then` as `choice` is to `or`
  - Closes #199
- Bump dependencies
  - `hashbrown`: 0.11 -> 0.12
  - `regex`: 1.5 -> 1.6
  - `pom`: 3.0 -> 3.2
  - `criterion`: 0.3.5 -> 0.4
- Also explicitly implement `Copy` + `Clone` for `Choice` to handle cases where the output type is not

### DIFF
--- a/src/zero_copy/primitive.rs
+++ b/src/zero_copy/primitive.rs
@@ -2,7 +2,7 @@ use super::*;
 
 pub struct End<I: ?Sized>(PhantomData<I>);
 
-pub fn end<I: Input + ?Sized>() -> End<I> {
+pub const fn end<I: Input + ?Sized>() -> End<I> {
     End(PhantomData)
 }
 
@@ -37,7 +37,7 @@ where
 
 pub struct Empty<I: ?Sized>(PhantomData<I>);
 
-pub fn empty<I: Input + ?Sized>() -> Empty<I> {
+pub const fn empty<I: Input + ?Sized>() -> Empty<I> {
     Empty(PhantomData)
 }
 
@@ -144,7 +144,7 @@ impl<T: Clone, I: ?Sized, E, S> Clone for Just<T, I, E, S> {
     }
 }
 
-pub fn just<T, I, E, S>(seq: T) -> Just<T, I, E, S>
+pub const fn just<T, I, E, S>(seq: T) -> Just<T, I, E, S>
 where
     I: Input + ?Sized,
     E: Error<I>,
@@ -206,7 +206,7 @@ impl<T: Clone, I: ?Sized, E, S> Clone for OneOf<T, I, E, S> {
     }
 }
 
-pub fn one_of<T, I, E, S>(seq: T) -> OneOf<T, I, E, S>
+pub const fn one_of<T, I, E, S>(seq: T) -> OneOf<T, I, E, S>
 where
     I: Input + ?Sized,
     E: Error<I>,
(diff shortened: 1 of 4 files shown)


src/zero_copy/primitive.rs
```diff
@@ -511,3 +511,81 @@ macro_rules! impl_choice_for_tuple {
...
+pub struct Group<T> {
+    parsers: T,
+}
+
+pub const fn group<T>(parsers: T) -> Group<T> {
+    Group { parsers }
+}
+
+// recursively combine to flatten a tuple
+macro_rules! recursive_combine {
```
wackbyte, line 526: Could this be simplified to a recursive `M::map`? I think I already tried that, but I'm not sure.

zesterer: Do you have a usage example of this?

wackbyte: I've added a test to try and demonstrate it

src/zero_copy/primitive.rs
```diff
@@ -511,3 +511,81 @@ macro_rules! impl_choice_for_tuple {
...
+        where
+            I: Input + ?Sized,
+            E: Error<I>,
+            S: 'a,
+            $($X: Parser<'a, I, E, S>),*
+        {
+            type Output = ($($X::Output,)*);
+
+            fn go<M: Mode>(&self, inp: &mut InputRef<'a, '_, I, E, S>) -> PResult<M, Self::Output, E> {
+                let before = inp.save();
```
wackbyte, line 572: Is this necessary? I noticed `then` doesn't rewind if the first parser fails and was wondering if it would be similar here.

zesterer: Yep, it's fine to skip the rewinding here. If a parser fails, it's not obliged to preserve the input state. It's up to parsers that care about recovering from failure (`.or(...)`, etc.) to preserve the state.

zesterer:
I think this in a sufficient place to merge, although I am still sceptical of its usefulness given that most then chains often include many occurrences of `then_ignore` and `ignore_then`.

PR #208, opened 2022-10-07
`Parser::then_not` for an equivalent of PEG `!`

mia-entropy:
When trying to convert a grammar I wrote in PEG into chumsky, I had a lot of trouble translating PEG `!`.

`Parser::not` which was added recently seems at first the solution, but it has the caveat of parsing a 1 input when the pattern fails to parse. This thus cause issues if we're at the end of the text, and require the usage of `then_ignore`, `rewind`, `end`, `or`, and even `map` to make to make the output of `end` `()` match the output of `not` `I`.

I thus suggest a new combinator `then_not` that wraps all of this, such that PEG `a !b` can be translated to `a.then_not(b)`.

Additionally, it might be useful for users to have a small section about the equivalence between PEG operators and chumsky combinators, as the documentation says "chumsky’s parsers are recursive descent parsers and are capable of parsing parsing expression grammars (PEGs)" but doesn't give more details.

### DIFF
--- a/src/combinator.rs
+++ b/src/combinator.rs
@@ -16,6 +16,10 @@ pub type IgnoreThen<A, B, O, U> = Map<Then<A, B>, fn((O, U)) -> U, (O, U)>;
 /// See [`Parser::then_ignore`].
 pub type ThenIgnore<A, B, O, U> = Map<Then<A, B>, fn((O, U)) -> O, (O, U)>;
 
+/// See [`Parser::then_not`].
+pub type ThenNot<A, B, I, O, U, E> =
+    ThenIgnore<A, Rewind<Or<Map<Not<B, U>, fn(I), I>, End<E>>>, O, ()>;
+
 /// See [`Parser::or`].
 #[must_use]
 #[derive(Copy, Clone)]
@@ -1343,7 +1347,7 @@ impl<I: Clone, O, A: Parser<I, O, Error = E>, F: Fn(E) -> Result<O, E>, E: Error
                 Ok(out) => {
                     stream.revert(start);
                     Ok((out, None))
-                },
+                }
             },
         };
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -792,6 +792,54 @@ pub trait Parser<I: Clone, O> {
         Map(Then(self, other), |(o, _)| o, PhantomData)
     }
 
+    /// Parse one thing only if it is not followed by another thing, yielding
+    /// the output of the former. This will also parse if the former is located at
+    /// the end of the output.
+    ///
+    /// The output type of this parser is `O`, the same as the original parser.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// # use chumsky::{prelude::*, error::Cheap};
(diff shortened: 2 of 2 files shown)


zesterer:
Hello, sorry I've not had the time to review this. Unfortunately I've got a lot on my plate at the moment, but I should have time this weekend coming.

It's true that the existing behaviour of `not` is perhaps a little unintuitive and could do with changing. It should probably accept End Of Input (unless the inner parser was itself `end()`), as you say. The current behaviour, if desired, could be recovered with `any().and_is(b.not())` instead of `b.not()`. I've been meaning to add an `and_is` combinator for a while, it turns out to be rather useful for a number of other patterns.

TL;DR, my proposal would be:

- Change the behaviour of the existing `a.not()` parser to make it accept EoI
- Add an `a.and_is(b)` parser that succeeds if both `a` and `b` succeed at parsing the current input (i.e: `"foo"` is parsed by both `any().repeated()` and `just("foo")`)

What do you think about this?

mia-entropy:
Yes, sounds good to me (with `a.not()` parser thus returning `()`). To be sure I understand correctly, `a.and_is(b)` would thus be equivalent to `empty().then_ignore(b).then(a).map(|(_, out)| out)` ?

By the name `and_is`, I also could expect that both `a` and `b` match the same span of input. `"fooa"` would thus be parsed entirely, which may be confusing. I had a similar surprise with `a.repeated().at_most(5)` that parsed even if `a` was repeated 6 times, leaving the 6th occurence unconsumed and needing additional code to forbid it. Maybe it is just me and the name is not that confusing, but that should probably be a remark in the docs?

By the way, I made [my own parser combinator library](https://github.com/eberban/eberban-rust/blob/main/crates/framework/src/lib.rs) inspired by yours since I didn't like how chumsky seems to discard errors in case another path succeeds. I went for a model that distinguishes failing to parse and encountering something forbidden. Feel free to take a look and grab stuff you like and think would be useful in chumsky :)

zesterer:
> To be sure I understand correctly, `a.and_is(b)` would thus be equivalent to `empty().then_ignore(b).then(a).map(|(_, out)| out)` ?

No, `a.and_is(b)` would be equivalent to `a.rewind().then(b).map(|(out, _)| out)`, so that it only succeeds if the same input is accepted by both parsers.

> I had a similar surprise with `a.repeated().at_most(5)` that parsed even if `a` was repeated 6 times, leaving the 6th occurence unconsumed and needing additional code to forbid it.

This is a good point, I'll make sure to document this.

> By the way, I made [my own parser combinator library](https://github.com/eberban/eberban-rust/blob/main/crates/framework/src/lib.rs) inspired by yours since I didn't like how chumsky seems to discard errors in case another path succeeds. I went for a model that distinguishes failing to parse and encountering something forbidden. Feel free to take a look and grab stuff you like and think would be useful in chumsky :)

I'll definitely take a look at it!

zesterer:
@mia-entropy You might be interested to know that #211 implements the changes discussed here, and in particular `not` is implemented as you suggested. This is only on the `zero-copy` branch for now, but once merged it shall be the default behaviour!

zesterer:
Closing given that this has effectively been resolved in `zero-copy`.

PR #211, opened 2022-10-27
zero-copy: add `and_is` and `not` combinators

wackbyte:
I've been putting this PR off for a little while out of concern for what I thought was strange error behavior:
```sh
# Using code from the "tree" example for `not`
> {{{{
[found '{' at 1..2]
> {{{{}
[found '{' at 1..2]
> {{{{}}
[found '{' at 1..2]
> {{{{}}}
[found end of input at 7..7]
# Wouldn't the first three make more sense if they were similar "end of input" errors?
```
However, this also happens when `none_of("{}")` is used instead of `one_of("{}").not()`, so it's not a problem unique to `not` itself.

I got the errors I expected on master:
```sh
> {{{{
found end of input but expected one of "{", "}"
> {{{{}
found end of input but expected one of "}", "{"
# ...etc
```
so I assume it may related to the differences in error handling between the two branches.

Also, I went ahead and ported `not`'s example and wrote one for `and_is`.

### DIFF
--- a/src/zero_copy/combinator.rs
+++ b/src/zero_copy/combinator.rs
@@ -971,7 +971,7 @@ where
     S: 'a,
     A: Parser<'a, I, E, S>,
 {
-    type Output = I::Token;
+    type Output = ();
 
     fn go<M: Mode>(&self, inp: &mut InputRef<'a, '_, I, E, S>) -> PResult<M, Self::Output, E> {
         let before = inp.save();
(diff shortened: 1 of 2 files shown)


zesterer:
Yep, the problem is likely that `zero-copy` does not yet properly implement error prioritisation.

src/zero_copy/combinator.rs
```diff
@@ -959,6 +959,96 @@ where
...
+            )),
+            Err(_) => {
+                if let Some(tok) = tok {
+                    Ok(M::bind(|| tok))
+                } else {
+                    Err(Located::at(
+                        at,
+                        E::expected_found(None, None, inp.span_since(before)),
+                    ))
+                }
```
zesterer, line 996: As discussed in #208, I think this case is probably more problematic than it's worth and `not` should instead parse no tokens at all on success.

zesterer:
Thanks! This is ideal :)

PR #212, opened 2022-11-01
update hashbrown (ahash)

### LINKED ISSUE #210: Outdated ahash dependency

zyansheep:
closes #210 

`cargo test` seem to run fine for me, not sure what MSRV chumsky uses tho...


zesterer:
The MSRV is likely to jump suddenly when `zero-copy` is merged because it relies on GATs anyway, so I don't think that's too much of an issue for this crate. Assuming the latest stable is probably fine.

PR #216, opened 2022-11-13
[TRY] Change zero_copy::Parser to take output type as generic type

bew:
Hello @zesterer,

This is a tentative at https://github.com/zesterer/chumsky/issues/9#issuecomment-1292788564 , indeed, it wasn't an easy change,

What I did basically (in a few more steps):
1. Change `zero_copy::Parser` to take `O` generic type, and remove its `Output` associated type. (start review here I think)
2. Go through all errors one by one and fix all combinators, primitives, etc... until `cargo build [...]` is happy.

   I've left a number of `FIXME`s, because I had the error `E0207` (a lot):
   For a code like this (simplified):
   ```rust
   impl<A, OA, ..., O> Parser<.., O, ..> for Map<A, F>
   where
       A: Parser<.., OA, ..>,
       F: Fn(OA) -> O,
       ...
   ```
   I get:
   ``the type parameter `OA` is not constrained by the impl trait, self type, or predicates``
   For some reason the compiler doesn't see that `OA` is used in `A`, which is used for `Map<A, F>` (Am I missing something? Is it a known limitation/bug(?) of the compiler?).
   To fix this I had to add a lot of additional PhantomData to combinator structs (like `Map<A, OA, F>`) to ensure `OA` (and others) are mentioned in either the impl trait or self type.

   Another thing: I didn't understand how to edit the macro `impl_group_for_tuple`, so it's not called for now (commented), I've commited what I thought would work, could you help me with that?

3. Try to run the tests, fix (AFAIK) the callers to use the parsers with the new `O` generic type.
4. => Now I'm blocked at an error I don't understand how to fix...:
```
error[E0599]: the method `or` exists for struct `zero_copy::combinator::MapSlice<zero_copy::combinator::Repeated<zero_copy::combinator::Filter<zero_copy::primitive::Any<_, _, _>, [closure@src/zero_copy/mod.rs:799:21: 799:31]>, char, _, (), _, _>, fn(&str) -> Token<'_> {Token::<'_>::Ident}, _, _>`, but its trait bounds were not satisfied
   --> src/zero_copy/mod.rs:810:14
    |
810 |             .or(string)
    |              ^^ method cannot be called on `zero_copy::combinator::MapSlice<zero_copy::combinator::Repeated<zero_copy::combinator::Filter<zero_copy::primitive::Any<_, _, _>, [closure@src/zero_copy/mod.rs:799:21: 799:31]>, char, _, (), _, _>, fn(&str) -> Token<'_> {Token::<'_>::Ident}, _, _>` due to unsatisfied trait bounds
    |
   ::: src/zero_copy/combinator.rs:5:1
    |
5   | pub struct MapSlice<A, F, E = (), S = ()> {
    | -----------------------------------------
    | |
    | method `or` not found for this struct
    | doesn't satisfy `_: zero_copy::Parser<'_, _, _, _, _>`
    |
note: trait bound `zero_copy::combinator::MapSlice<zero_copy::combinator::Repeated<zero_copy::combinator::Filter<zero_copy::primitive::Any<_, _, _>, [closure@src/zero_copy/mod.rs:799:21: 799:31]>, char, _, (), _, _>, fn(&str) -> Token<'_> {Token::<'_>::Ident}, _, _>: zero_copy::Parser<'_, _, _, _, _>` was not satisfied
   --> src/zero_copy/blanket.rs:5:8
    |
3   | impl<'a, T, I, O, E, S> Parser<'a, I, O, E, S> for &'a T
    |                         ----------------------     -----
4   | where
5   |     T: Parser<'a, I, O, E, S>,
    |        ^^^^^^^^^^^^^^^^^^^^^^ unsatisfied trait bound introduced here
note: the following trait must be implemented
   --> src/zero_copy/mod.rs:177:1
    |
177 | pub trait Parser<'a, I: Input + ?Sized, O, E: Error<I> = (), S: 'a = ()> {
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    = help: items from traits can only be used if the trait is implemented and in scope
note: `zero_copy::Parser` defines an item `or`, perhaps you need to implement it
   --> src/zero_copy/mod.rs:177:1
    |
177 | pub trait Parser<'a, I: Input + ?Sized, O, E: Error<I> = (), S: 'a = ()> {
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
(screenshot for colors)
image

This errors indicates that `zero_copy::Parser` is not implemented for `zero_copy::combinator::MapSlice`, but it is? The error mentions the blanket impl of `zero_copy::Parser` for all `&'a T`, but I don't understand why this is needed (there's not blanket in master branch afaik), and what is its relation to the error message..

---
I hope this work will be useful to you, at least to start rolling the ball for this, learn what not to do, .. :)

I'm happy to continue working on this if you manage to understand this error / unblock the situation, I learned a lot here!
Please tell me if you see any improvement to the structure of the changes, I'll propagate suggestions as needed.

### DIFF
--- a/examples/zero-copy.rs
+++ b/examples/zero-copy.rs
@@ -1,6 +1,6 @@
 use chumsky::zero_copy::prelude::*;
 
-fn parser() -> impl for<'a> Parser<'a, str, Rich<str>, Output = char> {
+fn parser() -> impl for<'a> Parser<'a, str, char, Rich<str>> {
     just('a').or(just('b'))
 }
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -13,8 +13,6 @@
 #![deny(missing_docs)]
 #![allow(deprecated)] // TODO: Don't allow this
 
-#![feature(associated_type_defaults)] // TODO: remove when removing 'type Output = O'
-
 extern crate alloc;
 
 pub mod chain;
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -13,6 +13,8 @@
 #![deny(missing_docs)]
 #![allow(deprecated)] // TODO: Don't allow this
 
+#![feature(associated_type_defaults)] // TODO: remove when removing 'type Output = O'
+
 extern crate alloc;
 
 pub mod chain;
--- a/src/zero_copy/blanket.rs
+++ b/src/zero_copy/blanket.rs
@@ -1,14 +1,12 @@
 use super::*;
 
-impl<'a, T, I, E, S> Parser<'a, I, E, S> for &'a T
+impl<'a, T, I, O, E, S> Parser<'a, I, O, E, S> for &'a T
 where
-    T: Parser<'a, I, E, S>,
+    T: Parser<'a, I, O, E, S>,
     I: Input + ?Sized,
     E: Error<I>,
     S: 'a,
 {
-    type Output = T::Output;
-
     fn go<M: Mode>(&self, inp: &mut InputRef<'a, '_, I, E, S>) -> PResult<M, Self::Output, E>
     where
         Self: Sized,
--- a/src/zero_copy/blanket.rs
+++ b/src/zero_copy/blanket.rs
@@ -7,12 +7,12 @@ where
     E: Error<I>,
     S: 'a,
 {
-    fn go<M: Mode>(&self, inp: &mut InputRef<'a, '_, I, E, S>) -> PResult<M, Self::Output, E>
+    fn go<M: Mode>(&self, inp: &mut InputRef<'a, '_, I, E, S>) -> PResult<M, O, E>
     where
         Self: Sized,
     {
         (*self).go::<M>(inp)
     }
 
-    go_extra!();
+    go_extra!(O);
 }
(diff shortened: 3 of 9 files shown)


wackbyte:
I've just made [this implementation of `impl_group_for_tuple`](https://gist.github.com/wackbyte/b00fd0bd28c7af192ab4825ff3e4fcbf) and I think it should work.

zesterer:
This is a great start!

> I get:
> the type parameter `OA` is not constrained by the impl trait, self type, or predicates
> For some reason the compiler doesn't see that OA is used in A, which is used for Map<A, F> (Am I missing something?

The reason for this error is that `OA` is only used in the trait bounds, not in the typing relationship itself. To solve this, you'll likely need to add an `OA` type parameter to the parser itself (backed by `PhantomData`).

If you wish, I can try to take this branch over when I next have time (probably in a few weeks).

herkhinah:
I've just fixed the `MapSlice` implementation in https://github.com/herkhinah/chumsky/tree/shekhinah/zero-copy-MapSlice

PR #222, opened 2022-11-25
Never box a parser twice

schneiderfelipe:
This just specializes `boxed` for `BoxedParser` so that it returns itself instead of boxing it again.


zesterer:
Great idea, thanks!

Issue #231, opened 2022-12-14
TODO in nano_rust


Timmmm:
In `nano_rust.rs` there this todo:

```rust
        block_chain
            // Expressions, chained by semicolons, are statements
            .or(raw_expr.clone())
            .then(just(Token::Ctrl(';')).ignore_then(expr.or_not()).repeated())
            .foldl(|a, b| {
                let span = a.1.clone(); // TODO: Not correct
                (
                    Expr::Then(
                        Box::new(a),
                        Box::new(match b {
                            Some(b) => b,
                            None => (Expr::Value(Value::Null), span.clone()),
                        }),
                    ),
                    span,
                )
            })
```

I was just curious why it isn't correct. Is it that the span is just from the first expression? Would the correct fix be this?

```rust
                let a_start = a.1.start;
                let b_end = b.as_ref().map(|b| b.1.end).unwrap_or(a.1.end);
                (
                    Expr::Then(
                        Box::new(a),
                        Box::new(match b {
                            Some(b) => b,
                            None => (Expr::Value(Value::Null), b_end..b_end),
                        }),
                    ),
                    a_start..b_end,
                )
```

In other words, the span for the whole expression is from the start of `a` to the end of `b` (or end of `a` if `b` is `None`), and then the span for `b` is an empty slice at the end of `a` if `b` is `None`.

Is that right?

Timmmm:
Awesome, thanks for the reply. Guess I may as well do a PR!

PR #232, opened 2022-12-14
Fix Expr::Then span in nano_rust example

Timmmm:
This sets the Expr::Then span correctly, resolving the only TODO in nano_rust.

Fixes TODO in nano_rust #231


zesterer:
Thanks! Any chance you could add a comment explaining that it's creating a span that converts the entire group of statements?

zesterer:
Much appreciated :)

PR #233, opened 2022-12-17
Add skip_parser recovery strategy

rumpuslabs:
I had a tricky little problem where I wanted an error recovery strategy to swallow everything up to a newline, but only if the line didn't contain a certain keyword which marked an alternative path that I wanted to take instead. I came up with this:

```rust
my_parser.recover_with(skip_parser(
    none_of([Token::Newline, Token::BadKeyword])
        .repeated()
        .then_ignore(just(Token::Newline)),
));
```

The implementation turned out very simple in the end and it would be extremely flexible (the other strategies could be implemented using it, although I imagine they would be a bit slower).

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -57,7 +57,7 @@ pub mod prelude {
             any, choice, empty, end, filter, filter_map, just, none_of, one_of, seq, take_until,
             todo,
         },
-        recovery::{nested_delimiters, skip_then_retry_until, skip_until},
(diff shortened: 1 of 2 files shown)


zesterer:
Thanks! This is, thereabouts, how I was intending to implement recovery strategies anyway on `zero-copy` (i.e: the strategy itself just looks for an alternative parse pattern) so this is very much in line with that theme for the time being.

rumpuslabs:
NB: I've just seen #77 and #107 which this replaces.

PR #238, opened 2022-12-31
Better label merging

rumpuslabs:
Sequences of `or()` parsers that all fail have their combined errors reported using the label of the first alternative in the list. Similarly, sequences of `then(xyz.or_not())` parsers have their combined errors reported using the last in the sequence.

I wanted these cases to be reported using none of their respective labels, so that a higher level label could be applied to their collective failure. However my first attempt to fix this just in `error::Simple::merge()` using something like `Option::xor` failed, because when processing three `.or()` branches in a row, the first two would have their label collapsed to `None` and then the third would replace the `None` with its own.

To address this issue I added a new type, `SimpleLabel`, which captures the difference between merging with an error with no label, or merging with an error that's already had multiple other labels merged into it.

While debugging this PR I found another issue with `.labelled("...")` branches seemingly not being applied correctly, it turns out that an error can sneak through the `Label<_, _>` parser as an "alternate error", so I've added a second commit to fix this.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -180,6 +180,36 @@ impl<I: fmt::Display, S: fmt::Display> fmt::Display for SimpleReason<I, S> {
     }
 }
 
+/// A type representing zero, one, or many labels applied to an error
+#[derive(Clone, Copy, Debug, PartialEq)]
+enum SimpleLabel {
+    Some(&'static str),
+    None,
+    Multi,
+}
+
+impl SimpleLabel {
+    fn merge(self, other: Self) -> Self {
+        match (self, other) {
+            (SimpleLabel::Some(a), SimpleLabel::Some(b)) if a == b => SimpleLabel::Some(a),
+            (SimpleLabel::Some(_), SimpleLabel::Some(_)) => SimpleLabel::Multi,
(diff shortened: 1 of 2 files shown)


src/error.rs
```diff
@@ -180,6 +180,35 @@ impl<I: fmt::Display, S: fmt::Display> fmt::Display for SimpleReason<I, S> {
...
+enum SimpleLabel {
+    Some(&'static str),
+    None,
+    Multi,
+}
+
+impl SimpleLabel {
+    fn merge(self, other: Self) -> Self {
+        match (self, other) {
+            (SimpleLabel::Some(_), SimpleLabel::Some(_)) => SimpleLabel::Multi,
```
zesterer, line 194: I think checking for label equality here is important: without that, two `"expression"` labels would get merged into `Multi`, which is probably not desirable.

zesterer:
Thanks, this seems like a broadly desirable change. Just one comment about the current implementation.

It's worth noting that when `zero-copy` is merged, I think I'll probably be moving things like labels into a series of non-default extension traits in order to keep the core of the crate as simple as possible.

zesterer:
Thanks!

PR #239, opened 2023-01-02
changes on top of bew [rebased]

herkhinah:
Because I wanted to use the zero-copy parser my self I implemented stuff on top of bew changes (rebased onto master)
- fixed the pythonic example
- changed the implementation of Ignored so it's less superfluous on trait bound requirements
- implemented `Seq` for `&[T]`
- added some text parsers (including semantic_indentation although one might change the implementation now that there are state full parsers)
- fixed the json bench example
- added Char trait methods from non-zero-copy Character
- fixed MapSlice
- fixed some doc code examples

I can undo the rebase, split the commit or do whatever. Also by any chance does anyone know when traits/trait bounds become superfluous and the compiler suddenly needs a lot of annotations?

For the text parsers I wasn't really sure what input types to use `C: Char` and `C::Slice` or `I: Input` and `Input::Token: Char` or something entirely else?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,7 +2,7 @@
 name = "chumsky"
 version = "0.8.0"
 description = "A parser library for humans with powerful error recovery"
-authors = ["Joshua Barretto <joshua.s.barretto@gmail.com>"]
+authors = ["Joshua Barretto <joshua.s.barretto@gmail.com>", "Elijah Hartvigsen <elijah.reed@hartvigsen.xyz"]
 repository = "https://github.com/zesterer/chumsky"
 license = "MIT"
 keywords = ["parser", "combinator", "token", "language", "syntax"]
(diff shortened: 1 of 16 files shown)


bew:
NOTE: You'll need to change the target branch of the PR to the `zero-copy` branch to only have relevant diffs

zesterer:
Argh, it's rather frustrating that this got rebased on to master because it means that the diff now includes *everything* in `zero-copy`: do you think you could cherry-pick your changes and those of @bew on to a new branch to avoid this problem?

herkhinah:
> Argh, it's rather frustrating that this got rebased on to master because it means that the diff now includes _everything_ in `zero-copy`: do you think you could cherry-pick your changes and those of @bew on to a new branch to avoid this problem?

Okay, I cherry picked the commits, now I'll just fixup the broken test cases.

PR #242, opened 2023-01-03
add the ability to have subslice parser combinators

herkhinah:
I want to be able to implement parser combinators where the second parser parses only the input slice the previous parser has parsed. E.g. an `AndIs` implementation where the second parser is limited to the input the first parser has parsed. 

As I'm making changes that feel more to the core than my previous I wanted to ask if I'm on the right track implementation wise and if such functionality would be welcome.

I've added a (badly named) trait `Slice` that is a `SliceInput` with the additional constraint `Self = SliceInput::Slice`. The use case is for parser combinators like the one described above so that both parser can have the same error type. Then there is the `with_subslice` function that limits the input in `InputRef` for the passed closure to the passed range.

### DIFF
--- a/src/zero_copy/input.rs
+++ b/src/zero_copy/input.rs
@@ -1,7 +1,9 @@
+use core::ops::AddAssign;
+
 use super::*;
 
 pub trait Input {
-    type Offset: Copy + Ord;
+    type Offset: Copy + Ord + AddAssign;
     type Token;
     type Span: Span;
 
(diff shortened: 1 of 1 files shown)


src/zero_copy/input.rs
```diff
@@ -161,6 +173,32 @@ pub struct InputRef<'a, 'parse, I: Input + ?Sized, E: Error<I>, S> {
...
 
+impl<'a, 'parse, I: Slice + ?Sized, E: Error<I>, S> InputRef<'a, 'parse, I, E, S> {
+    // used for parser combinators that want to temporarily shrink the input
+    pub(crate) fn with_subslice<T, F>(&mut self, range: Range<Marker<I>>, f: F) -> T where
+        F: Fn(&mut Self) -> T
+    {
+        let mut input = self.slice(range);
+        let mut marker = Marker {
+            pos: 0,
+            offset: input.start(),
```
zesterer, line 184: We need to be a bit careful here: `offset` is used for prioritising errors, and this effectively ends up lying to the rest of the crate about the offset of an error, which could have strange ramifications.

zesterer:
I think I'm a bit uncomfortable about merging this until we have a specific use-case for it (such as `AndIs`, as you say). It's likely that this code will need changing a little for such a case anyway, so I'd prefer to merge it along with code that actually uses it and hence can be tested.

herkhinah:
Yes, it was a hack for a use case that I don't have anymore.

PR #245, opened 2023-01-04
zero_copy::text cleanup

herkhinah:
I've forgot to add the state type parameter, so the `zero_copy::text` parsers can only be used with `()` as state. Also there where some unnecessary `Clone` trait bounds in `zero_copy::text::semantic_indentation`


zesterer:
Thanks!

PR #247, opened 2023-01-07
Start adding zero-copy docs

CraftSpider:
Adds documentation to all zero_copy items ~~And one or two minor changes I couldn't resist while documenting~~

### DIFF
--- a/src/zero_copy/primitive.rs
+++ b/src/zero_copy/primitive.rs
@@ -161,6 +161,16 @@ impl<'b, T: Clone, const N: usize> Seq<T> for &'b [T; N] {
     }
 }
 
+impl<'b, T: Clone> Seq<T> for Vec<T> {
+    type Iter<'a> = alloc::vec::IntoIter<T>
+    where
+        Self: 'a;
+
+    fn iter(&self) -> Self::Iter<'_> {
+        self.clone().into_iter()
+    }
+}
+
 impl Seq<char> for str {
     type Iter<'a> = core::str::Chars<'a>
     where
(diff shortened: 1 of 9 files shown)


src/zero_copy/primitive.rs
```diff
@@ -1,7 +1,57 @@
...
+///
+/// This parser is very useful when you wish to force a parser to consume *all* of the input. It is typically combined
+/// with [`Parser::then_ignore`].
+///
+/// The output type of this parser is `()`.
+///
+/// # Examples
+///
+/// ```
+/// # use chumsky::prelude::*;
```
bew, line 31: hmm shouldn't you be using the zero copy prelude instead?
```suggestion
/// # use chumsky::zero_copy::prelude::*;
```
(same for ALL examples)

CraftSpider: Correct, I haven't yet updated most of the examples. I'm currently working on documenting things new in zero_copy, then once that's done, I plan to do a massive search-and-replace before cleaning up all the errors that causes

CraftSpider:
Unrelated note: After this PR, definitely think it might be a good idea to pull `Container`, `Seq`, etc into their own module, to track these common 'container helpers'.

CraftSpider:
Currently `recover_with` fails, and two docs are TODO. On the other hand, I think it may be worth to merge now and fix those things later, since this change is very prone to bitrot

zesterer:
Do you think it's ready for merge?

CraftSpider:
I'm going to add at least basic documentation to the two TODO ones quick, then I think so

CraftSpider:
I'd call this PR ready to merge - there are one or two rough spots, but overall I think this improves the current state of zero-copy documentation.

zesterer:
Thanks! This is a big step forward :)

PR #251, opened 2023-01-09
[zero-copy] Make zero_copy work on stable

CraftSpider:
Preserves the `OnceCell` workaround from master, and adds a `MaybeUninit` extension for the new array combinators

### DIFF
--- a/src/zero_copy/combinator.rs
+++ b/src/zero_copy/combinator.rs
@@ -9,6 +9,28 @@ use super::*;
 use core::mem::MaybeUninit;
 use hashbrown::HashSet;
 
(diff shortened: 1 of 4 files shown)


src/zero_copy/combinator.rs
```diff
@@ -9,6 +9,28 @@ use super::*;
 use core::mem::MaybeUninit;
 use hashbrown::HashSet;
 
+// TODO: Remove this when MaybeUninit transforms to/from arrays stabilize in any form
+trait MaybeUninitExt<T>: Sized {
```
zesterer, line 13: Clever!

PR #252, opened 2023-01-09
[zero-copy] Fix recover_with doctest and bug it uncovered

CraftSpider:


CraftSpider:
Gonna ask forgiveness not permission, since this is basically just a one-line bugfix

PR #253, opened 2023-01-09
[zero-copy] Make parse return a custom type

CraftSpider:
I wasn't sure on this till after I changed all the tests - I personally it looks a lot nicer this way

### DIFF
--- a/src/zero_copy/error.rs
+++ b/src/zero_copy/error.rs
@@ -52,9 +52,9 @@ use super::*;
 ///     None => Err(MyError::NotADigit(span, c)),
 /// });
 ///
-/// assert_eq!(numeral.parse("3").0, Some(3));
-/// assert_eq!(numeral.parse("7").0, Some(7));
-/// assert_eq!(numeral.parse("f").1, vec![MyError::NotADigit(0..1, 'f')]);
+/// assert_eq!(numeral.parse("3"), ParseResult::Ok(3));
+/// assert_eq!(numeral.parse("7"), ParseResult::Ok(7));
+/// assert_eq!(numeral.parse("f").into_errors(), vec![MyError::NotADigit(0..1, 'f')]);
 /// ```
 pub trait Error<I: Input + ?Sized>: Sized {
     /// Create a new error describing a conflict between expected inputs and that which was actually found.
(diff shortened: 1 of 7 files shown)


src/zero_copy/mod.rs
```diff
@@ -74,6 +75,80 @@ use self::{
...
+/// will always have at least one element.
+#[derive(Debug, Clone, PartialEq)]
+pub enum ParseResult<T, E> {
+    /// Parser succeeded with no issues
+    Ok(T),
+    /// Parser succeeded with errors (hit a recovery path, for example)
+    Warn(T, Vec<E>),
+    /// Parser failed with errors
+    Err(Vec<E>),
+}
```
zesterer, line 88: Hmm, I'm unconvinced that an enum is the way to go. I think I'd prefer to see something like

```rs
struct ParseResult<T, E> {
    output: Option<T>,
    errors: Vec<E>,
}
```

That way, we can hide the implementation and add features like actual warnings (i.e: not errors) in the future without it being a breaking change.

src/zero_copy/error.rs
```diff
@@ -52,9 +52,9 @@ use super::*;
 ///     None => Err(MyError::NotADigit(span, c)),
 /// });
 ///
-/// assert_eq!(numeral.parse("3").0, Some(3));
-/// assert_eq!(numeral.parse("7").0, Some(7));
-/// assert_eq!(numeral.parse("f").1, vec![MyError::NotADigit(0..1, 'f')]);
+/// assert_eq!(numeral.parse("3"), ParseResult::Ok(3));
```
zesterer, line 55: I think it's probably worth using `.parse("3").into_result()` and then comparing with `Ok(...)` in the examples to minimise confusion.

CraftSpider:
Updated to a struct, and changed `into_tuple` into two methods - `into_result` which matches current `Parser::parse`, and `into_output_errors` which matches current `Parser::parse_verbose`, and which should have naming and behavior forward-compatible with future extensions to parser output

zesterer:
Thanks, that's ideal! Nice idea to do this by the way, definitely the superior API now.

PR #254, opened 2023-01-09
[zero-copy] Minor cleanups

CraftSpider:
Make benchmarks build again, fix doc link and change `ParseResult::errors` return type, add some inline to trivial functions


CraftSpider:
RE: The skip_while change

Unfortunately it looks worse. OTOH it also improves the JSON benchmark by ~%20 by my measurements, so I figured it was worth it

PR #255, opened 2023-01-10
[zero-copy] Update examples, and fix uncovered issues

CraftSpider:
This mostly fixes copy/clone bounds, along with a change to add some getters to `Rich`. This leaves some things commented out for features not yet fully ported to zero_copy.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,7 +33,7 @@ ariadne = "0.1.2"
 pom = "3.2"
 nom = "7.1"
 serde_json = { version = "1.0", features = ["preserve_order"] }
-criterion = "0.4.0"
+criterion = { version = "0.4.0", features = ["html_reports"] }
 
 [[bench]]
 name = "json"
--- a/examples/brainfuck.rs
+++ b/examples/brainfuck.rs
@@ -2,7 +2,7 @@
 //! Run it with the following command:
 //! cargo run --example brainfuck -- examples/sample.bf
 
-use chumsky::prelude::*;
+use chumsky::zero_copy::prelude::*;
 use std::{
     env, fs,
     io::{self, Read},
@@ -20,7 +20,7 @@ enum Instr {
     Loop(Vec<Self>),
 }
 
-fn parser() -> impl Parser<char, Vec<Instr>, Error = Simple<char>> {
+fn parser<'a>() -> impl Parser<'a, str, Vec<Instr>, Rich<str>, ()> {
     use Instr::*;
     recursive(|bf| {
         choice((
@@ -32,9 +32,10 @@ fn parser() -> impl Parser<char, Vec<Instr>, Error = Simple<char>> {
             just('.').to(Write),
         ))
         .or(bf.delimited_by(just('['), just(']')).map(Loop))
-        .recover_with(nested_delimiters('[', ']', [], |_| Invalid))
-        .recover_with(skip_then_retry_until([']']))
+        .recover_with(any().repeated().delimited_by(just('['), just(']')).to(Invalid))
+        .recover_with(take_until(just(']')).to(Invalid))
         .repeated()
+        .collect::<Vec<_>>()
     })
     .then_ignore(end())
 }
@@ -66,8 +67,8 @@ fn main() {
         .expect("Failed to read file");
 
     // let src = "[!]+";
-    match parser().parse(src.trim()) {
+    match parser().parse(src.trim()).into_result() {
         Ok(ast) => execute(&ast, &mut 0, &mut [0; TAPE_LEN]),
         Err(errs) => errs.into_iter().for_each(|e| println!("{:?}", e)),
-    }
+    };
 }
--- a/examples/json.rs
+++ b/examples/json.rs
@@ -3,7 +3,7 @@
 //! cargo run --example json -- examples/sample.json
 
 use ariadne::{Color, Fmt, Label, Report, ReportKind, Source};
-use chumsky::prelude::*;
+use chumsky::zero_copy::prelude::*;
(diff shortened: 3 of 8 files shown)


Cargo.toml
```diff
@@ -33,7 +33,7 @@ ariadne = "0.1.2"
 pom = "3.2"
 nom = "7.1"
 serde_json = { version = "1.0", features = ["preserve_order"] }
-criterion = "0.4.0"
+criterion = { version = "0.4.0", features = ["html_reports"] }
```
CraftSpider, line 36: I can revert this, it was accidentally left in, though also probably doesn't hurt

zesterer: It's testing only and IIRC it used to be enabled in criterion by default anyway, so I think it's fine.

examples/brainfuck.rs
```diff
@@ -32,9 +32,10 @@ fn parser() -> impl Parser<char, Vec<Instr>, Error = Simple<char>> {
             just('.').to(Write),
         ))
         .or(bf.delimited_by(just('['), just(']')).map(Loop))
-        .recover_with(nested_delimiters('[', ']', [], |_| Invalid))
-        .recover_with(skip_then_retry_until([']']))
+        .recover_with(any().repeated().delimited_by(just('['), just(']')).to(Invalid))
```
zesterer, line 35: It would be nice to have a few built-in recovery parsers akin to `nested_delimiters` and others.

Fwiw, this code is not equivalent to `nested_delimiters` because it doesn't account for nesting. For example, this will happily accept `[...[...]` as a recovery pattern, even though it should still be looking for a final `]`.

examples/json.rs
```diff
@@ -17,24 +17,25 @@ enum Json {
     Object(HashMap<String, Json>),
 }
 
-fn parser() -> impl Parser<char, Json, Error = Simple<char>> {
+fn parser<'a>() -> impl Parser<'a, str, Json, Rich<str>, ()> {
     recursive(|value| {
-        let frac = just('.').chain(text::digits(10));
+        let frac = just('.').then_ignore(text::digits(10))
+            .map_slice(|x| x);
```
zesterer, line 23: Is this a common enough operation that we could do with a `.slice()` combinator?

examples/json.rs
```diff
@@ -46,86 +47,83 @@ fn parser() -> impl Parser<char, Json, Error = Simple<char>> {
                 .or(just('r').to('\r'))
                 .or(just('t').to('\t'))
                 .or(just('u').ignore_then(
-                    filter(|c: &char| c.is_digit(16))
+                    any()
+                        .filter(|c: &char| c.is_digit(16))
                         .repeated()
```
zesterer, line 52: I wonder whether we should make `text::digits` return a `Repeated<impl Parser>` instead of an `impl Parser` so we can do `text::digits(16).exactly(4)` here instead.

examples/json.rs
```diff
@@ -46,86 +47,83 @@ fn parser() -> impl Parser<char, Json, Error = Simple<char>> {
...
                         .exactly(4)
-                        .collect::<String>()
-                        .validate(|digits, span, emit| {
+                        .map_slice(|x| x)
+                        .try_map(|digits, span| {
+                            char::from_u32(u32::from_str_radix(digits, 16).unwrap())
+                                .ok_or_else(|| Rich::expected_found(None, None, span))
+                        })
+                        .recover_with(empty().to('\u{FFFD}')) // unicode replacement character
```
zesterer, line 59: I don't think this works? If parsing `\uFFFF` or some other invalid character, this will reset the parser position to the `u` rather than the end of the thing.

zesterer:
Thanks! Just some comments about the examples :)

Cargo.toml
CraftSpider: Yeah, if you use the `.3` version, it gives some message along the lines of 'html_reports is becoming non-default, use it explicitly if you want it to keep working'. I added it in while playing with the benchmarks to get a better idea about the collected results.

examples/brainfuck.rs
CraftSpider: I knew it wasn't identical, but I was largely just getting things working. And I agree, lots of things already use `nested_delimiters` and it's probably useful to have some builtin strategies, even if they're no longer separate from other parsers. I'll see if I can port at least `nested_delimiters` and maybe some of the others, put them in a `recovery` module for now unless you think there's a better place.

examples/json.rs
CraftSpider: Probably, basically all the non-trivial examples use it at least once or twice (including nano_rust's non-pushed WIP port I have in my local, I'm still considering how best to port its spans)

examples/json.rs
CraftSpider: Hmm, you're right. I think we may be far enough along to actually implement `validate` now, since there's at least an error trait and at least minimal error prioritization again, so I might actually do that since that's really what this wants anyways.

examples/json.rs
CraftSpider: Hm, that might not be a bad idea. Though it may want its own type, unless we make that change to `repeated` to default into slice collecting on `I::Token`

CraftSpider:
I'm taking a quick pause on this PR - expect one expanding errors first, since I think that will make `nano_rust` easier and also avoid needing the commented-out blocks this PR has currently

examples/json.rs
zesterer: Tbh, I think not allocating by default for text parsers is the way to go where possible. We should probably do the same for `whitespace`, `int`, etc. too

zesterer:
I've ported these changes over to #275

PR #256, opened 2023-01-11
[zero-copy] Expand `Rich`, re-add `validate`

CraftSpider:


src/zero_copy/combinator.rs
```diff
@@ -1727,37 +1727,47 @@ where
...
-impl<'a, I, O, E, S, A, F> Parser<'a, I, O, E, S> for Validate<A, F>
+impl<'a, I, OA, U, E, S, A, F> Parser<'a, I, U, E, S> for Validate<A, OA, F>
 where
     I: Input + ?Sized,
     E: Error<I>,
     S: 'a,
-    A: Parser<'a, I, O, E, S>,
-    F: Fn(E, I::Span, &mut dyn FnMut(E)) -> E,
+    A: Parser<'a, I, OA, E, S>,
+    F: Fn(OA, I::Span, &mut dyn FnMut(E)) -> U,
```
zesterer, line 1754: I wonder whether we should get rid of the trait object in favour of a dedicated struct to avoid the dynamic dispatch... `Emitter<E>` and `emitter.emit(my_error)` :thinking:

CraftSpider: That doesn't sound horrible - and allows for future extensibility, if there's ever call for, say, custom error prioritization

src/zero_copy/error.rs
```diff
@@ -166,29 +326,27 @@ where
...
-            }
-        }
-        Ok(())
+        self.inner_fmt(f, I::Token::fmt, I::Span::fmt)
+    }
+}
+
+impl<I: Input + ?Sized> fmt::Display for Rich<I>
+where
+    I::Span: fmt::Display,
```
CraftSpider, line 335: This requirement is a little restrictive, since most spans are `Range` - opinions on whether I should change it to debug, or do something else?

zesterer: I think this is most correct, even if a little annoying. Perhaps we should provide our own `SimpleSpan` type? That way we can also add a `Copy` bound to it, as well as a `Display` impl.

src/zero_copy/error.rs
```diff
@@ -118,20 +118,169 @@ where
...
+                expected,
+                found,
+            } => {
+                write!(f, "found ")?;
+                write_token(f, token, &found)?;
+                write!(f, " at ")?;
+                span(&self.span, f)?;
+                write!(f, "expected ")?;
+                match &expected[..] {
+                    [] => write!(f, "something else")?,
```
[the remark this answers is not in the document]
CraftSpider: I think this is the same issue as with `not` and similar, where there's
A) Not a good way to tell how an arbitrary closure will expect tokens
B) Not a good way to get the 'inverse' of a set of tokens

The first I'm not sure how to fix generically, but the second we could maybe fix by making `expected_found` have a concept of 'inverted' sets, which may at least help users provide better errors in many cases, and allow a couple combinators to better express their rules

src/zero_copy/error.rs
CraftSpider: That sounds like a good idea. It would need to entirely replace `Range` - since `Span` is associated with `Input`, but I'm not entirely against that, especially as it's not too hard to make it pretty trivially convertable, and `zero_copy` breaks the world anyways.

src/zero_copy/error.rs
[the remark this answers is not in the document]
zesterer: Point A (and sort of point B) could be resolved by having the user provide a label like `.filter(|c| c.is_alphanumeric(), "alphanumeric")`, but that feels like quite a big thing to commit users to... or perhaps it's not? Regardless, you can already achieve the same with `.filter(|c| c.is_alphanumeric()).labelled("alphanumeric")`, so this doesn't feel like much of a win.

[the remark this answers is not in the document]
zesterer: Regardless, there's no need to resolve this in this PR of course.

CraftSpider:
Can remove the last commit if you want - but it was an idea I had, to make it more clear what the default error behavior is

src/zero_copy/span.rs
```diff
@@ -33,7 +33,58 @@ pub trait Span {
...
+        write!(f, "{:?}:{:?}", self.start, self.end)
+    }
+}
+
+impl<T> fmt::Display for SimpleSpan<T>
+where
+    T: fmt::Display,
+{
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{}:{}", self.start, self.end)
```
zesterer, line 83: Perhaps `{}..{}`? I'd associate `:` with `line:column`, which might be confusing in this case

zesterer:
Thanks!

PR #257, opened 2023-01-12
[zero-copy] Idea: repeat_prefixed for common then_with case

CraftSpider:
I consider `then_with` something of a chumsky 'worst-case' - it is sometimes needed, but I would much prefer as many cases as possible have alternatives. This PR implements one such alternative - the motivating case for me, length-prefixed parsing. I can't yet think of an amazing API for the more generic case to replace then_with, and I'm not 100% sure this is even a step in the right direction. But implementing it was easy, so I figured I'd put it up for comments

The `ThenWith` change is a drive-by, only parsers that do error handling should really be rewinding, doing so in `ThenWith` has a high chance to muck up error reporting.

### DIFF
--- a/src/zero_copy/combinator.rs
+++ b/src/zero_copy/combinator.rs
@@ -476,25 +476,9 @@ where
     F: Fn(OA) -> B,
 {
     fn go<M: Mode>(&self, inp: &mut InputRef<'a, '_, I, E, S>) -> PResult<M, OB, E> {
-        let before = inp.save();
-        match self.parser.go::<Emit>(inp) {
-            Ok(output) => {
-                let then = (self.then)(output);
-
-                let before = inp.save();
-                match then.go::<M>(inp) {
-                    Ok(output) => Ok(output),
-                    Err(e) => {
-                        inp.rewind(before);
-                        Err(e)
-                    }
-                }
-            }
-            Err(e) => {
-                inp.rewind(before);
-                Err(e)
-            }
-        }
+        let prefix = self.parser.go::<Emit>(inp)?;
+        let then = (self.then)(prefix);
+        then.go::<M>(inp)
     }
 
     go_extra!(OB);
--- a/src/zero_copy/primitive.rs
+++ b/src/zero_copy/primitive.rs
@@ -593,6 +593,70 @@ where
     go_extra!((C, OP));
 }
 
+/// See [`repeat_prefixed`]
+pub struct RepeatPrefixed<A, B, O, C = (), E = (), S = ()> {
+    prefix: A,
+    repeat: B,
+    phantom: PhantomData<(O, C, E, S)>,
+}
+
+impl<A, B, O, C, E, S> RepeatPrefixed<A, B, O, C, E, S> {
+    /// Set the type of [`Container`] to collect into.
+    pub fn collect<D: Container<O>>(self) -> RepeatPrefixed<A, B, O, D, E, S> {
+        RepeatPrefixed {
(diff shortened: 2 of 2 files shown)


zesterer:
I've been thinking about the issue of `then_with` and I've come to the conclusion that it's a hack to add context-sensitivity.

Context-sensitivity isn't necessarily a problem, but I think we need to come up with a more general framework for it. An idea I've been working on is something I call 'pass-forward' or 'pass-down' context sensitivity.

It would work something like this:

1) We'd add a new type parameter to `Parser` called `Ctx` or similar. It defaults to `()`. This context parameter is conceptually an *input* to a parser.

2) A combinator akin to `then_with` would be used like so:

```rs
// Let's imagine that collecting into `usize` just gives
// you the number of times the parser matched, for now
let indent = whitespace().collect::<usize>();

// This parser is configured by the context passed down
// to it. In this case, we say that it should accept exactly
// a certain number of whitespace characters based on whatever
// context is passed down to it.
let statements = whitespace()
    .configure(|cfg: &mut Repeated<...>::Config, n: usize| cfg.exactly(n))
    .ignore_then(expr)
    .then_ignore(newline())
    .repeated()

// Takes the output of `indent` and turns it into (via
// the closure) the context passed to `statements`
//
// The closure parameters are
// 1) The context that was passed to this parser from its
//    parent (there is no parent, so just `()` for now)
// 2) The output of the first parser (`indent`)
let block = indent.then_with_ctx(statements, |_parent_ctx: (), n: usize| n);
```

What do you think? This should fulfil the case you mention while also allowing for even more complex cases like identation sensitivity, as in the example.

CraftSpider:
I think that looks good in general - at a sketch, the new API sounds like this, if I'm right:
```rs
trait Parser<..., Ctx> {
    type Config;

    fn then_with_ctx(parser: impl Parser<...>, f: impl Fn(Ctx, O) -> NewCtx) -> ThenWithCtx<...>;

    fn configure(self, f: impl Fn(&mut Self::Config, Ctx)) -> Self;
}
```

If that looks right, I can try implementing it, instead of `RepeatPrefixed`

CraftSpider:
Wish there was a nice way to bundle up all these generics on `Parser` and by extension every combinator, but ah well

bew:
Could the Ctx be part of the State that the parser has already?

CraftSpider:
Quick answer: Maybe
Longer answer: I'd need to play around a bit, but I think that sounds feasible. Arguably the whole point of state is to add context while parsing, but currently it can't be used to actually effect it directly. On the other hand, making the current state the proposed context seems like it would be reasonable, and match with what I think when I hear 'parsing state'

CraftSpider:
Here's a question: Should `just` be configurable? The current example kind of needs it to be - you need to be able to set the expected value.

zesterer:
> Could the Ctx be part of the State that the parser has already?

I considered this... Yes, potentially: but you want it to have scoping rules, which `State` (deliberately) does not. i.e: parsers can only pass context *down* to their children and can't leak it up to parents or sideways to neighbours. You also don't want this context to be mutable either.

> Here's a question: Should just be configurable?

I think so, yes: but we can introduce configuration gradually, no need to jump in and make all parsers configurable yet. First we need to ensure that making them configurable doesn't have an unpleasant impact on performance.

CraftSpider:
I'm going to close this PR, as it looks like it's not the path forwards. I'll probably open another eventually for a context implementation, though that may take a bit - even without adding actual config, it's a lot of rote work.

PR #259, opened 2023-01-15
[zero-copy] Pull container into module, add more impls

CraftSpider:
We do a little bit of `Seq` optimization. This also lets us relax a couple bounds on some primitives. Overall, the change should be pretty minimal, but it should improve performance more the more expensive cloning the tokens/containers was. (benchmarks on my machine say this is only about a 1% win for json, but it does seem to be pretty consistently a little faster)

### DIFF
--- a/src/zero_copy/container.rs
+++ b/src/zero_copy/container.rs
@@ -125,15 +125,22 @@ impl<T, const N: usize> ContainerExactly<T, N> for [T; N] {
 /// This trait is likely to change in future versions of the crate, so avoid implementing it yourself.
 pub trait Seq<T> {
     /// The item yielded by the iterator
-    type Item<'a>: core::borrow::Borrow<T>
+    type Item<'a>: Borrow<T>
     where
         Self: 'a;
+
     /// An iterator over the items within this container, by reference.
     type Iter<'a>: Iterator<Item = Self::Item<'a>>
     where
         Self: 'a;
+
     /// Iterate over the elements of the container.
     fn seq_iter(&self) -> Self::Iter<'_>;
+
+    /// Check whether an item is contained within this sequence
+    fn contains(&self, val: &T) -> bool
+    where
+        T: PartialEq;
 }
 
 impl<T> Seq<T> for T {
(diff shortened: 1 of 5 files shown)


CraftSpider:
This apparently *regresses* `none_of`, possibly due to extra calls? I'm going to triage, see if it can be fixed, or if by value is just better for small tokens.

CraftSpider:
Well now I'm just confused, as the regression is gone. I think maybe my computer just spiked resources while the bench was running the first time.

CraftSpider:
`contains` is more of 'looks nicer' than a performance change. Overall I think this doesn't really change much performance-wise.

zesterer:
I have found that the json benchmark is *very* noisy.

src/zero_copy/container.rs
```diff
@@ -0,0 +1,456 @@
...
+        impl OrderedSeq<$ty> for core::ops::RangeInclusive<$ty> {}
+        )*
+    }
+}
+
+impl_for_range!(
+    u8, u16, u32, u64, usize,
+    i8, i16, i32, i64, isize,
+    char,
+);
```
zesterer, line 387: Can we not just generalise the impl to the bounds already on [`contains`](https://doc.rust-lang.org/std/ops/struct.Range.html#method.contains) ?

zesterer (commented):
Mostly looks good to me.

src/zero_copy/container.rs
CraftSpider: This is for the iterable part of ranges - they require `Step`, which is unstable. I could make `iter` part of `OrderedSeq`, so 'unordered' just lets you check bounds, then 'ordered' requires iteration, then it would just be a macro for the ordered parts. (I could also make them not ordered. That seems a bit disingenous though, as while they're not necessarily a common use case for `just`, they do have a strong order)

zesterer: That's unfortunate. Perhaps add a `// TODO: Generalise this with 'Step' later` to the code?

CraftSpider: The step issue provided me with a workaround I feel silly for not considering: `where Self: Iterator`

zesterer:
Thanks!

PR #260, opened 2023-01-20
[zero-copy] stabilize InputRef

CraftSpider:
Makes public the minimal API for custom parsers - no state access, non non-terminal errors, but you can parse, recover, and error correctly.

### DIFF
--- a/src/zero_copy/input.rs
+++ b/src/zero_copy/input.rs
@@ -198,10 +198,10 @@ impl<'a, 'parse, I: Input + ?Sized, E: Error<I>, S> InputRef<'a, 'parse, I, E, S
         }
     }
 
-    pub(crate) fn save(&mut self) -> Marker<I> {
+    pub fn save(&mut self) -> Marker<I> {
         self.marker
     }
-    pub(crate) fn rewind(&mut self, marker: Marker<I>) {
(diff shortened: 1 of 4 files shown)


src/zero_copy/input.rs
```diff
@@ -222,7 +222,7 @@ impl<'a, 'parse, I: Input + ?Sized, E: Error<I>, S> InputRef<'a, 'parse, I, E, S
         }
     }
 
-    pub(crate) fn next(&mut self) -> (usize, Option<I::Token>) {
+    pub fn next(&mut self) -> (usize, Option<I::Token>) {
```
zesterer, line 225: Part of me wonders whether we should stabilise the offset being returned. Perhaps we could have a `next_token` input that's just a light wrapper around `next` but drops the first field? What do you think?

CraftSpider: Currently the `usize` is used for `Located::at`, however, if we changed it to `Marker` we could drop the first field. I like the idea, it further reduces the chance of issues with user-facing API lock-in

zesterer: Sure! That seems like a sensible solution.

zesterer:
Ready to merge?

CraftSpider:
Yep

PR #261, opened 2023-01-20
WIP: [zero-copy] Implement `Ctx` and `Config` - parse-time control of parser parameters

CraftSpider:
The fundamental API principle here is that we add a new `Ctx` generic, and thread it through everything. Unlike state, it can only be passed down or forward along a parser stack, but it can in turn be used to configure future parsers with new behavior, through the `configure` combinator. Generics were renamed due to them becoming a bit of a trial to understand - and with the number of letters involved, a lot of places were starting to get confused about what `C` might mean.

This doesn't appear to detriment performance in non-configure cases very much, but allows context-sensitive parsing in a very nice-feeling way - no 'making new parsers', just 'using past information to configure behavior'.

The whole branch is highly WIP - tests have generally not been updated, and some combinators still don't have ctx threaded correctly.


CraftSpider:
One alternative way to handle the generic explosion is something like the following (this is just brainstorming):
```rs
trait Parser<I, O, Extra: ParserExtra = ExtraDefault> { ... }

trait ParserExtra {
    type Error;
    type State;
    type Ctx;
}

struct ExtraDefault;

impl ParserExtra for ExtraDefault {
    type Error = EmptyErr;
    type State = ();
    type Ctx = ();
}

struct ExtraError<E>(...)

impl<E> ParserExtra for ExtraError<E> {
    type Error = E;
}

struct ExtraFull<E, S, C>(...);

impl<E, S, C> for ExtraFull<E, S, C> {
    type Error = E;
    type State = S;
    type Ctx = C;
}
```

CraftSpider:
Whatever else can be said for the above idea - it sure does clean up a lot of generic signatures. Since the `E: Error<I>` and `S: 'a` bounds can be put on the trait, it lets most stuff stop caring about them.

bew:
Hello!
I can't comment on the meat of the PR, but I just wanted to say I like the renaming of generics :+1: 
If @zesterer is ok with it, I think it would be best to have that change in a separate PR though, to make it more isolated, and make this current PR easier to review as well (it will only have the changes specific to adding Ctx & Config systems).

CraftSpider:
That's probably a good idea, yeah. If they like #262, this one may have to wait a bit anyways - it would mean a very different implementation for this one, so I'd basically need to rework this anyways (though it would be faster to rework now than it was to first write)

CraftSpider:
I'm closing this due to large divergence - the generic rename can be its own thing if desired, and the rest is better built on top of the new ParserExtra

PR #262, opened 2023-01-20
[zero-copy] Replace `E: Error<I>, S: 'a` with `E: ParserExtra<'a, I>`

CraftSpider:
This implements the idea from #261 for improving generics in `Parser` - this may also help if, in the future, we want to split `Parser` into sub-crates for things like state and context, since then we could only require `Error` by default and add extra bounds for `State` or `Context`. It also cleans up a lot of generic signatures.

On the other hand, it's definitely a major change, and the naming may not be the best. I don't know if this is the desired path forwards, but I figured I'd put this up for opinions.

### DIFF
--- a/src/zero_copy/blanket.rs
+++ b/src/zero_copy/blanket.rs
@@ -1,13 +1,12 @@
 use super::*;
 
-impl<'a, T, I, O, E, S> Parser<'a, I, O, E, S> for &'a T
+impl<'a, T, I, O, E> Parser<'a, I, O, E> for &'a T
 where
-    T: Parser<'a, I, O, E, S>,
+    T: Parser<'a, I, O, E>,
     I: Input + ?Sized,
-    E: Error<I>,
-    S: 'a,
+    E: ParserExtra<'a, I>,
 {
-    fn go<M: Mode>(&self, inp: &mut InputRef<'a, '_, I, E, S>) -> PResult<M, O, E>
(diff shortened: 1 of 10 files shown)


CraftSpider:
Alternative to having `State` be a unique type is to have it be `type State<S> = ExtraFull<EmptyErr, S>`. There are some possible advantages to this in terms of all things with the same state and error always being the same type.

zesterer:
One thing I'm a little nervous about is how this could affect type inference. It doesn't exist yet, but I've been meaning to add a `.in_state(|state| &mut state.sub_state)` combinator that allows you to plug together two parsers with different state type parameters. Having the state be an associated type might harm the compiler's ability to flexibly infer state in such cases, but I'm not certain.

CraftSpider:
I think it would work fine, since I imagine that combinator would look something like `fn in_state<S>(self, f: impl FnOnce(E::State) -> S) -> InState<..., Full<E::Error, S>>` - just replace the extra param with a new one, and have the new state type follow normal generic inference. But I may try messing with it, see if it works right.

zesterer:
I had a bit more time to look over this, and I think I like this approach. It's a good way to shorten type signatures with move complexity out of the main body of the crate. Do you think it's ready?

CraftSpider:
I'll look over it again - I think I just drafted it since I wasn't sure whether it was the way forwards. I may make one more push to tighten up some stuff, I'll un-draft once it's ready for merge.

CraftSpider:
Changed everything to be typedefs over `Full`, since it's the same end effect, and I think it may help with future generic impls and such. I think this is ready.

Issue #265, opened 2023-02-02
[zero-copy] Way to foldr without allocating


CraftSpider:
Ran into this while talking with someone about converting to zero-copy - if you use `repeated` or similar, you can collect easily into nothing at all, or you can allocate. But there's a third case - where the user wants to just use the output as a number of times to fold. This can be re-written as a `slice(),map()` if the output is tokens, but if not, it's not super easy to work around - `Vec::<()>` might technically work, since ZST doesn't actually need to allocate, but not sure that many users will know or care about that, and it's still less efficient than just a running tally that can be iterated N times.

zesterer:
I think the easiest way to facilitate this would be to allow collecting into a `usize`.

zesterer:
I think I was half asleep when I read this the first time. I see what you're saying now.

There's another use-case that would likely be covered by whatever solution we find for this: parsers that generate many outputs rather than a single output (for example, it should really be possible to turn a lexer into an iterator of tokens).

zesterer:
Ideally we'd be able to do something akin to the following, for example:

```rs
let tokens = token.repeated();

for token_result in tokens.into_iter() {
    ...
}
```

CraftSpider:
Lazy parsing as a whole is an interesting concept, and would fit well with the iterator-like nature of the library. Might be interesting to look into, and would allow not collecting at all in many cases if one simply handles the intermediate values immediately. OTOH, I don't think it would be worth it if it requires too much of a break from existing style, but if it can be added without too much breakage, it would be quite nice.

zesterer:
I have an idea that should solve both issues, let me give it a go...

zesterer:
#270 should hopefully provide the tools we need to solve this problem and many similar problems.

zesterer:
This now happens for `repeated`/`separated_by`, but ironically not for `foldl`/`foldr`, the original subjects of the issue.

zesterer:
Closed as of #297. Unfortunately, it's not possible to avoid allocation for `foldr` but `foldl` is now allocation-free!

PR #266, opened 2023-02-03
[zero-copy] impl `Container` for `usize` + misc.

wackbyte:
Includes the `count` method, a shorthand for `.collect::<usize>()`.


wackbyte:
For #265. I've actually thought about adding this before, but I never got around to it.

zesterer:
Thanks! All looks good, much appreciated.

PR #269, opened 2023-02-08
[zero-copy] Implement `E::Context` and `Parser::Config` - parse-time control of parser parameters

CraftSpider:
Now with a more readable diff and less intrusive changes, thanks to `ParserExtra`

Only two configurations are provided so far - more may be added on-demand, but these two provide enough for useful examples and the most common use-case.

### DIFF
--- a/src/zero_copy/mod.rs
+++ b/src/zero_copy/mod.rs
@@ -493,7 +493,7 @@ pub trait Parser<'a, I: Input + ?Sized, O, E: ParserExtra<'a, I> = extra::Defaul
     fn configure<F>(self, cfg: F) -> Configure<Self, F>
     where
         Self: Sized,
-        F: Fn(Self::Config, &E::Context) -> Self::Config,
+        F: Fn(&mut Self::Config, &E::Context),
     {
         Configure {
             parser: self,
@@ -941,29 +941,58 @@ pub trait Parser<'a, I: Input + ?Sized, O, E: ParserExtra<'a, I> = extra::Defaul
     /// // A parser that parses a single letter and then its successor
     /// let successive_letters = one_of::<_, _, extra::Err<Simple<[u8]>>>(b'a'..=b'z')
     ///     .then_with_ctx(
-    ///         just(b'\0').configure(|cfg, ctx: &u8| cfg.set_seq(*ctx + 1)),
+    ///         just(b'\0').configure(|cfg, ctx: &u8| cfg.seq(*ctx + 1)),
     ///         |letter, _| letter,
     ///     );
     ///
     /// assert_eq!(successive_letters.parse(b"ab").into_result(), Ok(b'b')); // 'b' follows 'a'
     /// assert!(successive_letters.parse(b"ac").has_errors()); // 'c' does not follow 'a'
     /// ```
-    fn then_with_ctx<U, CtxN, P, F>(
+    fn then_with_ctx<U, CtxN, P>(
         self,
         then: P,
-        make_ctx: F,
-    ) -> ThenWithCtx<Self, P, O, F, I, extra::Full<E::Error, E::State, CtxN>>
+    ) -> ThenWithCtx<Self, P, O, I, extra::Full<E::Error, E::State, CtxN>>
     where
         Self: Sized,
         CtxN: 'a,
         P: Parser<'a, I, U, extra::Full<E::Error, E::State, CtxN>>,
-        F: Fn(O, &E::Context) -> CtxN,
+
     {
         ThenWithCtx {
             parser: self,
             then,
-            make_ctx,
             phantom: PhantomData,
+        }
+    }
+
+    /// TODO
+    fn map_ctx<Ctx, F>(
+        self,
+        mapper: F,
+    ) -> MapCtx<Self, F>
+    where
+        Self: Sized,
+        F: Fn(O, &E::Context) -> Ctx,
+        Ctx: 'a,
+    {
+        MapCtx {
+            parser: self,
+            mapper,
+        }
+    }
+
+    /// TODO
+    fn with_ctx<Ctx>(
+        self,
+        ctx: Ctx,
+    ) -> WithCtx<Self, Ctx>
+    where
+        Self: Sized,
+        Ctx: 'a + Clone,
+    {
+        WithCtx {
+            parser: self,
+            ctx,
         }
     }
 
--- a/src/zero_copy/mod.rs
+++ b/src/zero_copy/mod.rs
@@ -493,7 +493,7 @@ pub trait Parser<'a, I: Input + ?Sized, O, E: ParserExtra<'a, I> = extra::Defaul
     fn configure<F>(self, cfg: F) -> Configure<Self, F>
     where
         Self: Sized,
-        F: Fn(&mut Self::Config, &E::Context),
+        F: Fn(Self::Config, &E::Context) -> Self::Config,
     {
         Configure {
             parser: self,
--- a/src/zero_copy/mod.rs
+++ b/src/zero_copy/mod.rs
@@ -11,6 +11,12 @@ macro_rules! go_extra {
         fn go_check(&self, inp: &mut InputRef<'a, '_, I, E>) -> PResult<Check, $O, E::Error> {
             Parser::<I, $O, E>::go::<Check>(self, inp)
         }
+        fn go_emit_cfg(&self, inp: &mut InputRef<'a, '_, I, E>, cfg: Self::Config) -> PResult<Emit, $O, E::Error> {
+            Parser::<I, $O, E>::go_cfg::<Emit>(self, inp, cfg)
+        }
+        fn go_check_cfg(&self, inp: &mut InputRef<'a, '_, I, E>, cfg: Self::Config) -> PResult<Check, $O, E::Error> {
+            Parser::<I, $O, E>::go_cfg::<Check>(self, inp, cfg)
+        }
     };
 }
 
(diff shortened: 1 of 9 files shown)


zesterer:
Nice :)

Regarding the previous PR... Part of me regrets that the `Parser` trait has become more complex to use (even if easier to implement). I wonder whether it would be possible to keep the public API the same as before but have the utility of a secondary trait that hides the 'cruft' behind associated types, with a blanket impl. That said, I think the worst outcome would be for both traits to 'leak' into the public API, creating even more confusion.

src/zero_copy/mod.rs
```diff
@@ -423,6 +476,33 @@ pub trait Parser<'a, I: Input + ?Sized, O, E: ParserExtra<'a, I> = extra::Defaul
...
+    #[doc(hidden)]
+    fn go_emit_cfg(&self, inp: &mut InputRef<'a, '_, I, E>, cfg: Self::Config) -> PResult<Emit, O, E::Error>;
+    #[doc(hidden)]
+    fn go_check_cfg(&self, inp: &mut InputRef<'a, '_, I, E>, cfg: Self::Config) -> PResult<Check, O, E::Error>;
+
+    /// TODO
+    fn configure<F>(self, cfg: F) -> Configure<Self, F>
+    where
+        Self: Sized,
+        F: Fn(Self::Config, &E::Context) -> Self::Config,
```
zesterer, line 498: Perhaps `Fn(&mut Self::Config, &E::Context)`? LLVM sometimes has trouble optimising moves into references, even when inlining, and I don't see the value of passing by value really, especially if the config is unsized or something.

src/zero_copy/primitive.rs
```diff
@@ -116,6 +120,25 @@ where
...
+pub struct JustCfg<T> {
+    seq: Option<T>,
+}
+
+impl<T> JustCfg<T> {
+    /// TODO
+    pub fn set_seq(mut self, new_seq: T) -> Self {
+        self.seq = Some(new_seq);
+        self
+    }
```
zesterer, line 133: Maybe just `fn seq`/`fn sequence`? More consistent with `.exactly` etc. on other configs.

src/zero_copy/mod.rs
```diff
@@ -842,6 +922,49 @@ pub trait Parser<'a, I: Input + ?Sized, O, E: ParserExtra<'a, I> = extra::Defaul
...
+        P: Parser<'a, I, U, extra::Full<E::Error, E::State, CtxN>>,
+        F: Fn(O, &E::Context) -> CtxN,
+    {
+        ThenWithCtx {
+            parser: self,
+            then,
+            make_ctx,
+            phantom: PhantomData,
+        }
+    }
```
zesterer, line 966: Maybe `then_with_ctx` should lose the closure and use the first parser's output as the default context? Then we could have a separate `map_ctx` method if you want more control.

New example:

```rs
one_of::<_, _, extra::Err<Simple<[u8]>>>(b'a'..=b'z')
    .then_with_ctx(just(b'\0').configure(|cfg, ctx: &u8| cfg.set_seq(*ctx + 1)))
```

Python-style indentation-sensitive parsing:

```rs
let expr = just("expr"); // TODO

let block = recursive(|block| {
    let indent = text::whitespace()
        .configure(|cfg, parent_indent| cfg.at_least(*parent_indent))
        .count();

    let expr_stmt = expr.then_ignore(text::newline());
    let control_flow = just("loop:").then(text::newline()).ignore_then(block);
    let stmt = expr_stmt.or(control_flow);

    indent.then_with_ctx(stmt)
        .repeated()
        .map(Expr::Block)
});

block.with_ctx(0) // To give an initial context for everything
```

(parses the following)

```
block:
    expr
    block:
        expr
    expr
    block:
        expr
        block:
            expr
            expr
    expr
```

CraftSpider: I like this - we should probably have a separate context mapping combinator anyways, and this way it nicely 'separates concerns'

zesterer:
So I've been experimenting with this a little locally and I've hit a few issues.

```rs
#[derive(Clone, Debug)]
enum Stmt {
    Expr,
    Loop(Vec<Stmt>),
}

fn parser<'a>() -> impl Parser<'a, str, Vec<Stmt>> {
    let expr = just("expr"); // TODO

    let block = recursive(|block| {
        let indent = any().filter(|c| *c == ' ')
            .repeated()
            .configure(|cfg, parent_indent| {
                println!("Configure with {}", parent_indent);
                cfg.exactly(*parent_indent)
            });

        let expr_stmt = expr
            .then_ignore(text::newline())
            .to(Stmt::Expr);
        let control_flow = just("loop:")
            .then(text::newline())
            .ignore_then(block).map(Stmt::Loop);
        let stmt = expr_stmt.or(control_flow);

        text::whitespace().count()
            .then_with_ctx(stmt
                .separated_by(indent)
                .collect::<Vec<_>>(), |a, b| a)
    });

    // empty().then_with_ctx(block, |_: (), _: &()| 0)
    Parser::<_, _>::then_with_ctx(empty(), block, |_, _| 0)
}
```

First off: *this works!* Very cool stuff.

However, if I swap the last two lines (they're equivalent) I get the following compilation error:

```
error[E0284]: type annotations needed
  --> examples/indent.rs:34:13
   |
34 |     empty().then_with_ctx(block, |_: (), _: &()| 0)
   |             ^^^^^^^^^^^^^
   |
   = note: cannot satisfy `<_ as ParserExtra<'_, str>>::Error == EmptyErr`
help: try using a fully qualified path to specify the expected types
   |
34 |     <chumsky::zero_copy::primitive::Empty<str> as chumsky::zero_copy::Parser<'_, str, (), E>>::then_with_ctx::<Vec<Stmt>, usize, chumsky::zero_copy::recursive::Recursive<dyn chumsky::zero_copy::Parser<'_, str, Vec<Stmt>, chumsky::zero_copy::extra::Full<EmptyErr, (), usize>, Config = ()>>, [closure@examples/indent.rs:34:34: 34:49]>(empty(), block, |_: (), _: &()| 0)
   |     +++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++       ~

```

This seems to happen because the call to `.then_with_ctx` needs to use an implementation of `Parser`, but to do that it needs to figure out what `E: ParserExtra` is in use: but because we immediately override the context with a `usize` (0) there's nothing constraining this `E` so Rust can't infer anything.

I'm not sure how to resolve this. Perhaps it doesn't need solving? Ideally we'd have a `.with_ctx(0)` combinator that Avoids the need for the `empty().then_with_ctx` hack, but I'm fairly sure that it'll still end up having this constrained type variable floating around causing problems.

CraftSpider:
Forcing the context might work, but that does make issues with nested context-sensitive code. I couldn't do a `.then_with_context(array)` where `array = foo.then_with_context(value)`. Once there's `map_ctx` and possibly just `with_ctx`, they should help, but we may still need explicit ctx in some places. OTOH - this kind of inference failure is already an issue with state, and can be fairly easily fixed by adding `extra::Default` or `extra::Err<E>` somewhere on the preceding parser.

zesterer:
Agreed. Perhaps we should make `E` a type parameter of `fn empty` rather than just inferring it based on context.

CraftSpider:
You should now have access to `map_ctx` and `with_ctx`, which should help the issue

CraftSpider:
I can swap it back. That was kind of my idea behind doing it via passing ownership originally, it made the inline code nicer. I can swap it back, it's all of a three line change.

zesterer:
> I can swap it back. That was kind of my idea behind doing it via passing ownership originally, it made the inline code nicer. I can swap it back, it's all of a three line change.

I've already made the change in https://github.com/CraftSpider/chumsky/pull/1, feel free to merge it into your branch.

zesterer:
Oh, one other thing! This effectively replaces `then_with`, so I think it's worth getting rid of it given that it's opaque and potentially slow.

CraftSpider:
I wasn't sure whether we wanted to remove it yet, since a lot of parsers aren't yet configurable. OTOH, hopefully most of them will be by the time we release zero-copy, so I'm fine removing it now if you want.

CraftSpider:
Considering whether to split configuration into its own subtrait like IterParser - `CfgParser` or `CfgIterParser` which should help with the expanding `Parser` generics. OTOH, you still need `Context` on the extra, and it could lead to a bit of a trait explosion if we add more possible combinations :P

zesterer:
> Considering whether to split configuration into its own subtrait like IterParser - `CfgParser` or `CfgIterParser` which should help with the expanding `Parser` generics. OTOH, you still need `Context` on the extra, and it could lead to a bit of a trait explosion if we add more possible combinations :P

I think this seems sensible. I see `Context` as being sort of like `State`, something that 'moves through' parsers and hence makes sense to be on every parser. `Config`, however, is specific to each parser and independent so I think your idea of having a dedicated `ConfigParser` trait is a good one and should reduce overall complexity.

CraftSpider:
Update: Been busy the past couple days, will get this rebased soon. I'm most of the way through rewriting it into a new trait.

zesterer:
Ready to merge?

CraftSpider:
I would say so, if you think it looks good

PR #270, opened 2023-02-08
Added support for iterable parsers

zesterer:
Very simple implementation for now, doesn't handle failure. It's annoying that the implementation for parsers returning `impl IntoIterator` doesn't work due to an unconstrained type parameter, but hey ho.


zesterer:
@CraftSpider Out of interest, what are your thoughts on this?

CraftSpider:
I haven't gone too in-depth into it, but it looks reasonable at a first pass. My one question is whether this can be used to build any interesting combinators - can you write a combinator that `folds` on an iterable parser? Or one that collects it? (I'm not necessarily against it if you can't, but I'm curious how well this plays with existing combinators)

zesterer:
Actually, there is one solution that might work nicely:

Instead of `.collect()` being a method on `Repeated`/`SeparatedBy`, it becomes a method on `impl IterParser`. That solves the problem of the default 'phantom' collection type of `()`/`Empty` while also allowing us to have `Foldl`/`Foldr` just ingest `impl IterParser` without ever worrying about expensive intermediates being produced.

CraftSpider:
Oh, that sounds nice. `collect` turns an `IterParser` into a `Parser` with the collected value, and that type becomes irrelevant to `fold`

zesterer:
I'm going to take a crack at implementing this.

zesterer:
@CraftSpider I think I'm going to remove `chain`. In `master` it was effectively a hack to account for the fact that we couldn't use `map_slice`, but now we can: I don't think there's a good reason to provide the functionality given that:

- 80% of cases are covered by `map_slice`
- all others can be implemented manually anyway
- It's slow and tends to require a lot of inline allocation

zesterer:
@CraftSpider Ok, I think this is ready for review if you had a moment. Also happy to rebase on your context PR, your choice.

src/zero_copy/mod.rs
```diff
@@ -1676,6 +1610,174 @@ pub trait Parser<'a, I: Input + ?Sized, O, E: ParserExtra<'a, I> = extra::Defaul
...
+    fn next<M: Mode>(
+        &self,
+        inp: &mut InputRef<'a, '_, I, E>,
+        state: &mut Self::IterState<M>,
+    ) -> Option<PResult<M, O, E::Error>>;
+
+    /// Collect this iterable parser into a [`Container`].
+    ///
+    /// This is commonly useful for collecting parsers that many values outputs into containers of various kinds:
+    /// [`Vec`]s, [`String`]s, or even [`HashMap`]s. This method is analgous to [`Iterator::collect`].
```
CraftSpider, line 1665:
```suggestion
    /// [`Vec`]s, [`String`]s, or even [`HashMap`]s. This method is analagous to [`Iterator::collect`].
```

CraftSpider (approved):
I'm guessing SeparatedBy will become an `IterParser` in the future

CraftSpider:
Forgot to respond to other part of your request: I'll rebase on this once it's merged

src/zero_copy/mod.rs
zesterer: Thanks!

zesterer:
> I'm guessing SeparatedBy will become an `IterParser` in the future

Yes. I don't want to refactor that monster in this PR though :D

PR #277, opened 2023-02-13
[zero-copy] Make SeparatedBy an IterParser

CraftSpider:


zesterer:
Huh, impressive reduction in complexity! I assume the tests in `combinator.rs` still all pass?

CraftSpider:
They do, they caught a bug before the final version. We should probably start running those on CI for zero copy soon

PR #282, opened 2023-02-16
Perf

zesterer:


CraftSpider (approved):
We do a little bit of inlining - looks good to me, at a first pass

PR #283, opened 2023-02-16
[zero-copy] Add basic docs, fix up ctx

CraftSpider:
Tests pass locally, minus the regex stuff fixed in my CI PR


zesterer:
Happy to merge this, although I still feel like there's unfinished business with `map_ctx` :thinking:

PR #286, opened 2023-02-17
Add a Chain impl

figsoda:
It would be nice to unseal Chain, but this will work for my use case

### DIFF
--- a/src/chain.rs
+++ b/src/chain.rs
@@ -19,6 +19,7 @@ mod private {
     impl<T> Sealed<T> for Vec<T> {}
     impl<T> Sealed<T> for Option<Vec<T>> {}
     impl<T> Sealed<T> for Vec<Option<T>> {}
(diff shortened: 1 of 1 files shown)


zesterer:
Happy to merge this, although be aware that `chain` will be removed in the next major release (the `map_slice` combinator pretty much covers most use-cases and is much more efficient).

figsoda:
`map_slice` and the zero-copy branch looks really promising.
On an unrelated note, is `one_of("ab")` more efficient than `just('a').or(just('b'))`? ([this change](https://github.com/zesterer/chumsky/commit/0b26351ec6de8ebeb1e4bc3a83b106071aaaf4eb#diff-b200c3d75fd71695a4ee53cf4dc28b19b5a425c4b607a9fdcfe42d7495888b03R28) for example)

zesterer:
I think in most cases the compiler will tend to generate pretty much the same code (maybe `one_of` is slightly faster, but I think it depends on context). I just prefer `one_of` for ergonomic reasons.

PR #288, opened 2023-02-18
`zero-copy`: Add pratt parsing

Zij-IT:
### What is this PR?

This PR is a near 1:1 implementation of #144, updated to match `zero-copy` with the change that `pratt` is now function of the `Parser` trait (as seen in the second commit)

### API Concerns

In #144 @zesterer expressed their concerns regarding the mismatched API between this function and the rest of the crate. For this reason I went and did some brainstorming for a nice middle-ground between flexibility and matching the API of the crate. Here is a quick rundown of what I have in mind. Please let me know if/how it fits and if it is something that I should invest more time in implementing. This is currently just a rough draft, and as such I didn't spend a lot of time working out the generics.

#### User Side

The `Parser::pratt` function would now have a similar API to `Choice`, but instead of a tuple, the user would supply an array of `PrattOp` structs:

```rust
atom.pratt([
    PrattOp::new_left(0, just('+'), Expr::Add),
    PrattOp::new_left(0, just('-'), Expr::Sub),
    PrattOp::new_right(1, just('*'), Expr::Mul),
    PrattOp::new_right(1, just('/'), Expr::Div),
])
```

This `PrattOp` struct would be used to internalize the logic of `InfixPrecedence` as well as removing the necessity for `InfixOperator` and the `Operator` enum used in the examples provided by the original author of #144. `InfixPrecedence` and `Associativity` would, just like `Strength`, become internal implementation details and *not* be exposed to the user.

#### Developer Side

On `chumsky`'s side the implementation does get a little trickier, but I think it's  manageable. The `[PrattOp; N]` passed via the `Parser::pratt` function would be stored in a `PrattOperatorParser`.

```rust

pub struct PrattOp<I, O, E, Expr, P = u8> {
    prec: InfixPrecedence<P>,
    parser: Parser<I, O, E>,
    build: Fn(l: Expr, r: Expr) -> Expr
}

pub(crate) struct PrattOperatorParser<N: usize, I, O, E, Expr, P = u8> { 
    ops: [PrattOp<I, O, E, Expr, P>; N],
}
```

In the implementation of `Parser` for `PrattOperatorParser` instead of returning the operator it returns the `InfixPrecedence` and `build` function of the successful `PrattOp`. This is a small improvement over the current situation, because right now the operator parser must emit the operator, in order to get the precedence and the build function. Here is an example of how it's implementation could look like:

```rust
impl Parser for OpParser {
    fn go(&self, inp: &mut InputRef) -> PResult<Emit, (InfixPrecedence, F), E::Error> {
        let before = inp.save();
        let mut error: Option<Located<_> = None;

        for op in &self.ops {
            match op.parser.go::<Check>() {
                Ok(_) => return Ok(op.prec, op.build.clone()),
                Err(e) => {
                    err = Some(match err {
                        Some(err) => err.prioritize(e, |a, b| a.merge(b)),
                        None => e,
                    });
                    inp.rewind(before);
                }
            }
        }

        Err(err.unwrap_or_else(|| Located::at(inp.offset().into(), E::Error::expected_found(None, None, inp.span_since(before.offset)))))
    }
}
```

#### Final Remarks

With this change user's would have less boilerplate in writing their own Pratt-parsers, and the parser could be expanded in the future to include suffix and prefix operators via functions like `PrattOp::suffix` and `PrattOp::prefix`.

Do you feel that this would fit the rest of the API? Do you feel like this is something that could be of use in the current `zero-copy` branch? What concerns do you have about the idea or what problems could you see in its implementation?

Thanks for your time!

### DIFF
--- a/src/zero_copy/pratt.rs
+++ b/src/zero_copy/pratt.rs
@@ -144,121 +144,12 @@ pub trait InfixOperator<Expr> {
     fn build_expression(self, left: Expr, right: Expr) -> Expr;
 }
 
-/// See [`pratt()`].
+/// See [`Parser::pratt`].
 #[derive(Copy, Clone)]
 pub struct Pratt<Extra, AtomParser, Expr, OpParser, Op> {
     pub(crate) parser_op: OpParser,
     pub(crate) parser_atom: AtomParser,
     pub(crate) phantom: PhantomData<(Extra, Expr, Op)>,
-}
-
-/// Use Pratt parsing to efficiently parse binary operators
-/// with different associativity.
-///
-/// The parsing algorithm currently uses recursion
-/// to parse nested expressions.
-///
-/// # Examples
-///
-/// ```
-/// use chumsky::zero_copy::prelude::*;
-/// use chumsky::zero_copy::pratt::{pratt, InfixOperator, InfixPrecedence, Associativity};
-///
-/// #[derive(Clone, Copy, Debug)]
-/// enum Operator {
-///     Add,
-///     Sub,
-///     Mul,
-///     Div,
-/// }
-///
-/// enum Expr {
-///     Literal(i64),
-///     Add(Box<Expr>, Box<Expr>),
-///     Sub(Box<Expr>, Box<Expr>),
-///     Mul(Box<Expr>, Box<Expr>),
-///     Div(Box<Expr>, Box<Expr>),
-/// }
-///
-/// impl std::fmt::Display for Expr {
-///     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-///         match self {
-///             Self::Literal(literal) => write!(f, "{literal}"),
-///             Self::Add(left, right) => write!(f, "({left} + {right})"),
-///             Self::Sub(left, right) => write!(f, "({left} - {right})"),
-///             Self::Mul(left, right) => write!(f, "({left} * {right})"),
-///             Self::Div(left, right) => write!(f, "({left} / {right})"),
-///         }
-///     }
-/// }
-///
-/// impl InfixOperator<Expr> for Operator {
-///     type Strength = u8;
-///
-///     fn precedence(&self) -> InfixPrecedence<Self::Strength> {
-///         // NOTE: Usually, in Rust for example, all these operators
-///         // are left-associative. However, in this example we define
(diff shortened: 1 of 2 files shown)


src/zero_copy/pratt.rs
```diff
@@ -0,0 +1,363 @@
...
+    E: ParserExtra<'a, I>,
+    AtomParser: Parser<'a, I, Expr, E>,
+    OpParser: Parser<'a, I, Op, E>,
+    Op: InfixOperator<Expr>,
+{
+    fn go<M: Mode>(&self, inp: &mut InputRef<'a, '_, I, E>) -> PResult<M, Expr, E::Error>
+    where
+        Self: Sized,
+    {
+        self.pratt_parse::<M, _>(inp, None)
```
zesterer, line 203: Would it make sense to just inline `pratt_parse` here? I can't think of a reason to call it outside of this context anyway given that it already needs all of the 'mucky internals' associated with parsing (like `InputRef`).

zesterer (commented):
This is really great, thanks so much for porting it across! I do have a few minor reservations about the implementation, but I think they can be solved without too much difficulty.

Firstly, the current implementation uses a trait to specify precedence. I think it would be nicer if this could be done inline, as you say in your description.

Secondly, you mention a pattern like `atom.pratt([...])`, but an array would require either all elements to be of the same type, or would require boxing. Just as with `choice` and `group`, we can get around that by implementing the required trait with [tuples + a macro](https://github.com/zesterer/chumsky/blob/zero-copy/src/zero_copy/primitive.rs#L764-L827).

Other than that, I'm really excited to see this become a thing people can work with, it's definitely more ergonomic than the precedence climbing that's currently required!

src/zero_copy/pratt.rs
Zij-IT: I may be missing something obvious, but due to it's recursive nature requiring the `None` on the first call, and `Some(prec.strength_right())` on following calls I don't see how that's possible to inline the body into `Parser::go`. Unless you literally just mean adding the `#[inline]` tag, in which case that's not a problem :D

zesterer: Ah, no problem, I'm being a fool

Zij-IT:
Closing this in light of my most recent PR

Issue #74, opened 2022-01-15
Exponential blowup in recursive parsers


p-e-w:
Consider this simple parser:

```rust
use chumsky::prelude::*;

fn parser() -> impl Parser<char, String, Error = Simple<char>> {
  recursive(|expr| {
    let atom = text::ident()
      .or(expr.clone().delimited_by('(', ')'));

    let expression = atom
      .clone()
      .then_ignore(just('+'))
      .then(atom.clone())
      .map(|(a, b)| format!("{}{}", a, b))
      .or(atom);

    expression
  }).then_ignore(end())
}

fn main() {
  println!("{:?}", parser().parse("((((((((((((((((((((((a+b))))))))))))))))))))))"));
}
```

Parsing the string

```
((((((((((((((((((((((a+b))))))))))))))))))))))
```

takes **8 seconds** when compiling in release mode (around a minute in debug mode), and that time rapidly grows as more pairs of parentheses are added.

When the parser inside the `recursive` closure is more complex, the problem becomes even worse. With [the parser I'm working on](https://github.com/p-e-w/savage/blob/64b69c4395f4b9dfa850acbf30af9136966ffdd9/savage_core/src/parse.rs), I'm seeing multi-second parse times already for 3-4 levels of nested brackets.

Boxing parsers does not solve this issue. What can be done here?

zesterer:
The problem here is subtle, but it actually a more general problem with [backtracking](https://en.wikipedia.org/wiki/Backtracking) parsers, rather than an issue with chumsky itself. A similar problem will appear with a hand-written recursive descent parser.

Within your code, you have two patterns that can be parsed that both look *very similar*. They are:

```
expr
```

and

```
expr + expr
```

Notice something important: both patterns are ambiguous right the way up until the point at which the `+` is hit, if it exists. This means that chumsky (which is a recursive descent [PEG](https://en.wikipedia.org/wiki/Parsing_expression_grammar) parser internally) is forced to try one of the patterns first, then the next pattern if the first fails.

Normally, this is not a problem. For an expression like `a + b`, it's obvious - very quickly - that the second pattern is correct after just the second symbol. What really causes this to become a problem is that these patterns are defined recursively: at each level of recursion, the parser needs to try one pattern and then the other. If it guesses wrong the first time, it's just done a lot of work for nothing!

[...]

Issue #84, opened 2022-02-09
Implement memoisation


zesterer:
[Memoisation](https://en.wikipedia.org/wiki/Memoization#Parsers) is a technique for making backtracking parser performance linear over the input length at the cost of higher memory consumption and more overhead for less exponential grammars. It would be nice to be able to support it.

See also #74

zesterer:
Closing because #290 implements a memoisation combinator. However, I still believe there's a need for improved behaviour: it can be unintuitive to figure out where best to place the combinator right now.

PR #290, opened 2023-02-18
Added memoised parser

zesterer:


zesterer:
Closes #74

Zij-IT:
And #84 :)

Cargo.toml
```diff
@@ -14,13 +14,15 @@ exclude = [
...
-default = ["std", "spill-stack"]
+default = ["std", "spill-stack", "memoisation"]
 # Integrate with the standard library
 std = []
 # Enable nightly-only features like better compiler diagnostics
 nightly = []
 # Allows deeper recursion by dynamically spilling stack state on to the heap
 spill-stack = ["stacker", "std"]
+# Allows parser memoisation, speeding up heavily back-tracking parsers and allowing left recursion
+memoisation = []
```
zesterer, line 25: Maybe we should use the US spelling? Or a different feature name?

CraftSpider: I'm trying to think of a different name, and can't think of one. I personally don't feel strongly about the spelling, but I wish we could maybe just have two features with the different spellings that require each-other :P

Issue #182, opened 2022-08-10
Implement Parser for ! (never)


zesterer:
This will allow the use of `todo!()` in places that expect a parser.

msrd0:
I don't see how this would replace your existing parser? `todo!()` panics as soon as it's being called, i.e. when constructing the parser, while your `todo` parser (if I understood it correctly) only panics when it's used to parse something, not when it's being constructed.

zesterer:
This is true, it wouldn't be a substitute. It would be nice to have it just compile though. Thanks for pointing this out!

dvdvgt:
Hi! I really enjoyed using chumsky. I found the documentation really helpful, especially the tutorial on creating a little programming language. Thus, I'd like to contribute to this project. May I take this issue on? Is it still relevant?

zesterer:
Yes, this would be a good thing to implement in `zero-copy` rather than `master`. One problem is that in `zero-copy`, the output type is an associated type which might preclude implementing this.

zesterer:
`zero-copy` no longer has associated output types, so this should be possible (behind a nightly feature flag, of course).

CraftSpider:
Kind of sad that basically this alone prevents output from being associated - OTOH, I don't see a better option without language support. I've had at least one impl that would have been allowed with associated types that I couldn't write, but ah well.

zesterer:
Closing since #294 implemented this.

PR #294, opened 2023-02-21
[zero-copy] impl Parser for !

wackbyte:
For #182


zesterer:
Thanks!

wackbyte:
Note that `!` will not be able to coerce to `impl Parser` yet without `#![feature(never_type_fallback)]`.

PR #295, opened 2023-02-22
[WIP] [zero-copy] Static analysis

CraftSpider:
This is all slightly cursed, and horribly WIP, but I wanted it up for viewing

### DIFF
--- a/src/visit.rs
+++ b/src/visit.rs
@@ -246,12 +246,16 @@ where
 impl<A, O, C> Visitable for CollectExactly<A, O, C>
 where
     A: Visitable,
+    C: ContainerExactly<O>,
 {
     fn info(&self) -> ParserInfo<'_, Self> {
         ParserInfo::new(
             self,
             "collect_exactly",
-            todo!(),
+            SizeHint::repeat(
+                self.parser.info().size_hint,
+                SizeHint::new(C::LEN, Some(C::LEN))
+            ),
         )
     }
 


zesterer:
This is neat! I like the trait-based API. I feel like this is worth putting behind a feature flag though.

CraftSpider:
Yeah, probably. I'll add the feature flag, and try to flesh out more of the parsers with implementations. I'm still not *super* happy with how much state the visitor has to carry around, but I'm not sure how to do it much better while still being generic over any possible parser.

m-hugo:
Why is zesterer:zero-copy still targeted instead of zesterer:main ? it makes the diff unreadable

CraftSpider:
By mistake - fixed the target

zesterer:
Is this ready for review?

m-hugo:
why not expose this and pratt in an `experimental` (`use chumsky::experimental::{problems, pratt}`) module ?
that way it's not in indefinite limbo and you can rework it later

CraftSpider:
Once I have the motivation and time, I'll prioritise improving the API here, to be clear. It's just in limbo right now because I've both been busy and struggling to draw up the motivation to work on it.

zesterer:
> Once I have the motivation and time, I'll prioritise improving the API here, to be clear. It's just in limbo right now because I've both been busy and struggling to draw up the motivation to work on it.

That's ok! I find my motivation comes and goes a lot too. It definitely shows in my commit history :D

PR #314, opened 2023-03-02
Added stack spilling

zesterer:


soywod:
Looks like `stacker` does not build on `aarch64-*` and `windows` systems. Would it make sense to enable the feature only for `x86_64-linux` and `x86_64-darwin` systems?

soywod:
Yes, it is what I do from my side. I can definitely propose a PR, I will try to do that as soon as I can.

soywod:
The compilation should not fail, according to `stacker`:

> On all unsupported platforms this library is a noop. It should compile and run, but it won't actually grow the stack and code will continue to hit the guard pages typically in place.

It's been a while I did not try the feature back. If it still fails, then I will open a PR.

soywod:
Looks like it does not fail anymore, so we can discard the feature!

<https://github.com/pimalaya/mml/actions/runs/12569288083>

zesterer:
Nice, it's great to an issue that becomes a non-issue without any intervention required!

PR #316, opened 2023-03-03
Make ThenWithCtx and Map implement IterParser when relevant

CraftSpider:


src/util.rs
```diff
@@ -0,0 +1,76 @@
+use core::ops::{Deref, DerefMut};
+
+pub enum MaybeRef<'a, T> {
+    Ref(&'a T),
+    Own(T),
+}
```
zesterer, line 6: So I actually thought of a way we can remove the need for this last night by slightly adjusting iter parsers. I think this is probably not needed now. Other than that, this PR looks good.

CraftSpider:
I can drop the `MaybeMut` changes if you want - I think it looks a little nicer, but it's not a functional change

zesterer (approved):

PR #322, opened 2023-03-03
Removed state result from InputRef

zesterer:


CraftSpider (approved):
Approved, modulo the build passing

PR #324, opened 2023-03-03
bench(json): Include winnow in the cross-parser compare

epage:
`winnow` is a recent fork of `nom` that is being used by `tonl_edit` / `toml`.

`winnow` seems to be taking half the time of `nom` and I had to check
the output to confirm there wasn't a bug.  I dumped the `JsonZero`
structure from both parsers and diffed it and it was the same.

On my machine
```
json_nom
                        time:   [385.34 µs 386.74 µs 388.21 µs]
json_winnow
                        time:   [160.89 µs 161.85 µs 162.91 µs]
json_chumsky_zero_copy
                        time:   [265.28 µs 266.16 µs 267.06 µs]
```
These results are a complete surprise to me with `winnow` being the fastest parser combinator library, coming in at less than half the parse time  `nom`.  For `toml_edit`, `winnow` takes only 20% less time which I **think** I narrowed down to an extra `#[inline]`.  In this case, the main difference that I can think of is the use of `escaped` as its rewritten in `winnow` and `toml_edit` doesn't use it.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,6 +34,7 @@ regex = { version = "1.7", optional = true }
 ariadne = "0.1.5"
 pom = "3.2"
 nom = "7.1"
+winnow = "0.3"
 serde_json = { version = "1.0", features = ["preserve_order"] }
 criterion = "0.4.0"
 pest = "2.5"
--- a/benches/json.rs
+++ b/benches/json.rs
@@ -31,6 +31,10 @@ fn bench_json(c: &mut Criterion) {
         move |b| b.iter(|| black_box(nom::json(black_box(JSON)).unwrap()))
     });
 
+    c.bench_function("json_winnow", {
+        move |b| b.iter(|| black_box(winnow::json(black_box(JSON)).unwrap()))
+    });
+
     c.bench_function("json_chumsky_zero_copy", {
         use ::chumsky::prelude::*;
         let json = chumsky_zero_copy::json();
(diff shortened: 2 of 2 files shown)


zesterer:
Wow, looks like we have a new competitor to take on! Thanks for adding it to the benchmark.

epage:
@zesterer its still declarative, its mapping a parser to parsers :).  On a more serious note, we have it documented as a standard thing to use.  I held off on porting other optimizations from `toml_edit` like parsing strings in two modes: unescaped and escaped, as the parsers are written more naively so I figured I should be consistent.

Plus, `nom` and `winnow` are not meant to exclusively be used declarative.  [That was one of the big selling points for me to use `nom` and make `winnow`](https://epage.github.io/blog/2023/02/winnow-toml-edit-combine-nom/).

Also, the `dispatch` only helped by about 10%.

zesterer:
Interesting, perhaps a fast path for escaping should be next on our list to investigate. I've just taken a look at winnow, that's a great project you have there. I like the customisability of the thing, that's very nice. I'm hopeful that in time the slightly more closed down approach of chumsky will pay dividends, but I'm excited to see just how far the two approaches can be pushed. You've actually given me a few ideas for potential optimisations internally...

epage:
I suspect we can never match a fully optimized closed solution, and thats okay!  Some, like me, are willing to pay the cost for not feeling so constrained while for others, the closed solution will provide great results.

PR #337, opened 2023-03-06
Add CBOR benchmark

CraftSpider:
Still needs impls for other major libraries, but I can say current perf is quite promising:
image


src/error.rs
```diff
@@ -89,6 +89,13 @@ pub trait Error<'a, I: Input<'a>>: Sized {
 #[derive(PartialEq, Eq, PartialOrd, Ord, Debug, Copy, Clone)]
 pub struct EmptyErr(());
 
+impl EmptyErr {
+    /// TODO
```
CraftSpider, line 93: Feel free to make me actually document things in PRs as I go - I have a habit of just making changes then going back later, but that's not a great habit for a library this big probably :P

zesterer: I was fine with it provided there is a `TODO`: throw them everywhere you can, they're easy to search for.

zesterer: Maybe it's better to just have `EmptyErr` implement `Default`, by the way?

src/lib.rs
```diff
@@ -2137,6 +2137,20 @@ pub trait ConfigIterParser<'a, I: Input<'a>, O, E: ParserExtra<'a, I> = extra::D
             phantom: PhantomData,
         }
     }
+
+    /// A combinator that allows fallible configuration of the parser from the current context -
+    /// if an error is returned, parsing fails.
+    fn try_configure<F>(self, cfg: F) -> TryIterConfigure<Self, F, O>
```
zesterer, line 2143: Could this behaviour be replicated with `try_map` on the output of the parser that provides configuration? Or is this for a sub-branch of the context-consuming parser?

benches/cbor.rs
```diff
@@ -0,0 +1,221 @@
...
+
+            let float_simple = choice((
+                simple(20).to(CborZero::Bool(false)),
+                simple(21).to(CborZero::Bool(true)),
+                simple(22).to(CborZero::Null),
+                simple(23).to(CborZero::Undef),
+                simple(26).ignore_then(any()
+                    .repeated_exactly::<4>()
+                    .collect::<_, _, [_; 4]>()
+                    .map(f32::from_be_bytes)
```
zesterer, line 190: Out of interest, is this type hint necessary? It would be unfortunate if it was given the call to `map` just below.

src/error.rs
CraftSpider: That's probably a good idea, yeah

benches/cbor.rs
CraftSpider: They're not, they were before the map was added, but now it appears rustc has enough info to figure it out

src/lib.rs
CraftSpider: You can replicate it that way, yeah. It has a bit of the filter then map issue though - you validate the input, then need to either have an `unreachable` in the configuration, or map it to another type which can't have the relevant states.

zesterer (approved):

PR #348, opened 2023-03-07
Various example and API improvements

zesterer:


src/util.rs
```diff
@@ -1,36 +1,108 @@
...
     }
 }
+
+impl<'a, T> From<T> for Maybe<T, &'a T> {
+    fn from(x: T) -> Self {
+        Self::Val(x)
+    }
+}
+
+impl<'a, T> From<&'a T> for Maybe<T, &'a T> {
```
CraftSpider, line 104: Is it worth adding these for mut refs too?

zesterer: Will do

PR #350, opened 2023-03-07
Fix map-err combinators

CraftSpider:


CraftSpider:
Failure due to... ariadne not having a v1.6, just a v2? Interesting, as it didn't fail for your PR changing it.

zesterer:
Yes, I had to yank 0.1.6, didn't realise it contained a semver break. `2.0.0` should work fine though.

zesterer (approved):

PR #352, opened 2023-03-07
Replace `repeated_exactly` and `separated_by_exactly` with `collect_exactly`

CraftSpider:
CBOR benchmark doesn't show a notable regression, so that's an added plus

### DIFF
--- a/src/combinator.rs
+++ b/src/combinator.rs
@@ -6,7 +6,6 @@
 //! when accessed through their respective methods on [`Parser`].
 
 use super::*;
-use core::mem::MaybeUninit;
 
 /// The type of a lazy parser.
 pub type Lazy<'a, A, I, E> =
@@ -1161,8 +1160,7 @@ where
         }
     }
 
-    /// Require that the pattern appear exactly the given number of times. If the value provided
-    /// is constant, consider instead using [`Parser::repeated_exactly`]
+    /// Require that the pattern appear exactly the given number of times.
     ///
     /// ```
     /// # use chumsky::prelude::*;
@@ -1395,8 +1393,7 @@ where
         }
     }
 
-    /// Require that the pattern appear exactly the given number of times. If the value provided is
-    /// constant, consider instead using [`Parser::separated_by_exactly`].
+    /// Require that the pattern appear exactly the given number of times.
     ///
     /// ```
     /// # use chumsky::prelude::*;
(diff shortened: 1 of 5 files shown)


src/combinator.rs
```diff
@@ -1618,6 +1615,61 @@ where
...
+        let before = inp.offset();
+        let mut output = M::bind(|| C::uninit());
+        let mut iter_state = self.parser.make_iter::<M>(inp)?;
+        for idx in 0..N {
+            match self.parser.next::<M>(inp, &mut iter_state) {
+                Ok(Some(out)) => {
+                    M::combine_mut(&mut output, out, |c, out| C::write(c, idx, out));
+                }
+                Ok(None) => {
+                    inp.add_alt(
```
CraftSpider, line 1652: Would be nice to only do this if there's no error? I don't know

src/combinator.rs
```diff
@@ -1618,6 +1615,61 @@ where
     go_extra!(C);
 }
 
+/// See [`IterParser::collect_exactly`]
+pub struct CollectExactly<A, O, C, const N: usize> {
```
zesterer, line 1619: I wonder whether the `N` should become an associated type of `ContainerExactly` instead of a type parameter? Then we don't need to have this `N` hanging around in the type signatures when really it's already implied by the collection type

CraftSpider: We'd have to remove the impl for `()`, though that should already be covered by `repeated().exactly()`

zesterer: I think it's worth it to do that, personally

src/combinator.rs
zesterer: Another of these "we need to generate an error but we don't know how" cases :(

src/combinator.rs
```diff
@@ -1618,6 +1615,61 @@ where
...
+                        inp.offset,
+                        None,
+                        None,
+                        unsafe { inp.span_since(before) },
+                    );
+                    M::map(output, |mut output| unsafe { C::drop_before(&mut output, idx) });
+                    return Err(())
+                }
+                Err(()) => {
+                    M::map(output, |mut output| unsafe { C::drop_before(&mut output, idx) });
```
zesterer, line 1662: Maybe a `map_mut` that takes `&mut output` (like `combine_mut`) would speed up this case?

src/container.rs
```diff
@@ -153,6 +154,70 @@ impl<T, const N: usize> ContainerExactly<T, N> for [T; N] {
...
+    unsafe fn drop_before(uninit: &mut Self::Uninit, i: usize) {
+        // SAFETY: We're the only owners of the uninit data at this point
+        C::drop_before(unsafe { &mut *uninit.get() }, i)
+    }
+    unsafe fn take(uninit: Self::Uninit) -> Self {
+        Rc::from_raw(Rc::into_raw(uninit) as *mut C)
+    }
+}
+
+unsafe impl<T, C> ContainerExactly<T> for Arc<C>
```
zesterer, line 199: I guess we should eventually add similar impls for `Container`, but no need in this PR

zesterer (approved):

PR #365, opened 2023-03-10
Cleanup/Fix several TODO items

CraftSpider:


src/combinator.rs
```diff
@@ -1693,7 +1692,6 @@ pub struct OrNot<A> {
     pub(crate) parser: A,
 }
 
-// TODO: Maybe implement `IterParser` too?
```
CraftSpider, line 1696: There doesn't appear to be a reasonable way to implement `IterParser` for `OrNot` - if you have one in mind I can try that instead.

zesterer (approved):

src/lib.rs
```diff
@@ -1597,12 +1640,6 @@ where
     I: Input<'a>,
     E: ParserExtra<'a, I>,
 {
-    // TODO: Unfortunately we can't constrain the trait above without a cycle. Figure out a way to have this associated
```
CraftSpider, line 1600: The type is already usable from external crates a `P::Config` - sealed traits associated types can still be accessed (may want to make sure this doesn't cause any issues)

src/lib.rs
```diff
@@ -2016,7 +2052,6 @@ macro_rules! select {
 /// Useful if you want to extract elements from a token in a zero-copy manner.
 ///
 /// `select_ref` requires that the parser input implements [`BorrowInput`].
-// TODO: Remove this, somehow unify with `select`?
```
CraftSpider, line 2019: Since there doesn't seem to be a better way after discussion, figured this TODO can go away

src/combinator.rs
zesterer: That's fair enough

src/lib.rs
zesterer: Ah, nice. That's good to know.

Unfortunately, it doesn't appear in docs :(

src/lib.rs
zesterer: Yep, that's reasonable

src/lib.rs
CraftSpider: Unfortunate. May be worth adding a section on it at some point

Issue #303, opened 2023-02-28
Make `recursive` detect left recursion with debug assertions are enabled


zesterer:
This can be done by using similar machinery to the `memoised` combinator.

zesterer:
In addition, repeating combinators should detect when they don't make progress and panic.

PR #369, opened 2023-03-11
Detect left recursive parsers making no progress

CreatorSiSo:
Implementation of #303
Adds a debug panic for left recursive parsers without memoization.


zesterer:
Unfortunately, I don't think the changes in this PR are sufficient by themselves. For example, I'm pretty sure any mutually recursive pattern will not be caught, such as this:

```rs
recursive(|a| {
    recursive(move |_| {
        a
    })
})
```

I think unfortunately something more substantial is required, like piggy-backing on top of the memoisation hashmap.

PR #371, opened 2023-03-13
Left recursive checks

zesterer:


m-hugo:
false positive? this https://github.com/m-hugo/hugoc/tree/f39632f5f9d1d1bb658d051764cf38eabd2b3018
panics on debug but runs fine on release (except for a surprising `found end of input expected something else` before the actual end of the file)
slapping a dozen memoized in other places doesn't remove that `Recursive parser defined at src/main.rs:43:9 is left-recursive. Consider using .memoized() or restructuring this parser to be right-recursive.`

zesterer:
> false positive?

Yep, this is why this isn't merged yet. The check assumes that parsers are functionally pure, but `.memoized()` messes with this assumption for implementation reasons. I've yet to come up with a reliable and efficient way around it. When I do, I'll try to get this merged.

Issue #188, opened 2022-08-12
Use `track_caller` to allow panicking parsers to generate better error messages


Lokathor:
When the `todo()` parser panics it gives panic locations like
```
C:\Users\Daniel\.cargo\registry\src\github.com-1ecc6299db9ec823\chumsky-0.8.0\src\primitive.rs:841:9
```
And that's never helpful.

However, adding `track_caller` to just the `parse_inner` method probably won't do. It needs to be connected through at all layered steps in the parsing web.

CraftSpider:
The panic message should already include the phrasing 'Parser defined at <location>', at least on master. I remember trying to add the mentioned track_caller when looking into that, and it doesn't really help - even with a whole bunch of them, it would just point to where you called `parse`, not where the parser was defined. The panic message including the location was the best I could manage.

Lokathor:
Yeah i saw the one other closed issue but couldn't quite make sense of what had happened or if it was related to Todo / todo stuff.

I guess if it's fixed on github then zesterer just needs to get with the releasing.

CraftSpider:
I think there was hope that GATs would be stable soon, so there could be one major release (the breaking zero_copy changes), instead of one on the current stuff then shortly after zero-copy, but then they stalled a bit and stuff got dragged out. On the other hand, I'm not Zesterer, so he can say if I'm spreading misinformation on the internet.

zesterer:
I think we'll probably want to do one more release before `zero-copy`, but I've not done so yet because I don't want to find that GAT stabilisation is actually still 6 months off, leading to the two branches diverging (in terms of features) even more.

It's looking like we might get confirmation of stabilisation very soon, in which case I'll run a final `0.9` release of the current `master`, and then focus entirely on `zero-copy`.

zesterer:
Generalising this issue somewhat since I think it's a useful thing to do for more than just `todo` (`.unwrapped` could do with it, for example).

Zij-IT:
Looks like this was solved with #377

Do we want to keep this open?

zesterer:
No, this is resolved.

PR #377, opened 2023-03-14
Use user source location in more panic messages

CreatorSiSo:
This should cover the missing cases mentioned in #188


zesterer:
Thanks! This is ideal.

PR #380, opened 2023-03-17
Zero copy pratt parser

zesterer:
Ongoing to work to update #288.

Yet to do: reworking API to act a little nicer when used inline within a parser.

### DIFF
--- a/src/pratt.rs
+++ b/src/pratt.rs
@@ -144,121 +144,12 @@ pub trait InfixOperator<Expr> {
     fn build_expression(self, left: Expr, right: Expr) -> Expr;
 }
 
-/// See [`pratt()`].
+/// See [`Parser::pratt`].
 #[derive(Copy, Clone)]
 pub struct Pratt<Extra, AtomParser, Expr, OpParser, Op> {
     pub(crate) parser_op: OpParser,
     pub(crate) parser_atom: AtomParser,
     pub(crate) phantom: PhantomData<(Extra, Expr, Op)>,
-}
-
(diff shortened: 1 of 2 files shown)


src/pratt.rs
```diff
@@ -0,0 +1,371 @@
...
+            let (op, prec) = match self.parser_op.go::<Emit>(inp) {
+                Ok(op) => {
+                    let prec = op.precedence();
+                    if prec.strength_left().is_lt(&min_strength) {
+                        inp.rewind(pre_op);
+                        return Ok(left);
+                    }
+                    (op, prec)
+                }
+                Err(_) => return Ok(left),
```
TGMM, line 181: Just a heads up, I think this should be:
```
Err(_) => {
    inp.rewind(pre_op);
    return Ok(left);
}
```
Otherwise, things like `2+3R` where R is any random character will still parse and consume the character, preventing it from being used in future parsers.

Zij-IT:
Since I am determined to make sure this makes it in for the first `zero-copy` full release, is there anything in particular that needs to happen to it before it gets merged?

I know you mentioned taking a crack at making the API more inline with the rest of the crate. Had any luck in that regard?

zesterer:
> I know you mentioned taking a crack at making the API more inline with the rest of the crate. Had any luck in that regard?

Yep, this is the main thing at this stage. I've not made much progress on this unfortunately. I'm also a little bit concerned at what looks like an over-use of recursion in the implementation, but this is something that can be looked into without a change to the API.

Zij-IT:
closable?

PR #383, opened 2023-03-18
Add foldl_with_state and foldr_with_state

MechSlayer:


zesterer (commented):
Thanks! The implementation looks great. Is there any chance you could put something like `// TODO: Add examples of interning/arena allocation` below both of the doc comments for `Parser::fold[l/r]_with_state`?

MechSlayer:
Done, also changed the examples to since they were invalid @zesterer

zesterer:
Seems like inference is failing for the tests, I imagine you need to specify the missing type parameter.

MechSlayer:
My bad, still pretty new to Rust, didn't know that docs actually get compiled and act as tests

PR #385, opened 2023-03-25
tring → string (typo fix)

gimbles:


gimbles:
The.. check seems unrelated to my changes?

zesterer:
Thanks! Looks like it might need a fmt pass or something?

m-hugo:
> Thanks! Looks like it might need a fmt pass or something?

?, it's a clippy error, since https://github.com/rust-lang/rust-clippy/pull/10520 it thinks src/private.rs:280:18 is UB
280 |         unsafe { MaybeUninit::uninit().assume_init() }

zesterer:
> it thinks src/private.rs:280:18 is UB

That's frustrating. It would be nice if clippy checked the self type.

zesterer:
This is now fixed on main, so after a rebase CI should pass.

gimbles:
Merged, should pass CI now

PR #387, opened 2023-03-26
Ignore clippy lint

wackbyte:
For #385


zesterer:
Ah, I've already fixed this in 2be5cd. Thanks though!

PR #410, opened 2023-05-03
Use new `cargo-semver-checks-action`

mgr0dzicki:
Hi! As one of the contributors of `cargo-semver-checks` I'd want to say we're pleased to see you're using our tool!

Since we've recently released a new, completely rebuilt version of the GitHub action, you can now use it instead of installing `cargo-semver-checks` manually in CI. The new action handles the setup of Rust toolchain, downloads prebuilt binary of the latest `cargo-semver-checks` instead of building it from source and caches the baseline rustdoc between the runs. These all results in huge speedup of the semver job - less than 1m 30s compared to 6 minutes before.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -66,6 +66,6 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       - name: Check semver compatibility
-        uses: obi1kenobi/cargo-semver-checks-action@main
+        uses: obi1kenobi/cargo-semver-checks-action@v2
         with:
           rust-toolchain: nightly
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -65,11 +65,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
-      - name: Install latest nightly
-        uses: dtolnay/rust-toolchain@master
-        with:
-            toolchain: nightly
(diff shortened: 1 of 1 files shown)


zesterer:
Thanks, I appreciate you taking the time to open this PR!

zesterer:
@mgr0dzicki We've started getting some [ci failures](https://github.com/zesterer/chumsky/actions/runs/4967195915/jobs/8889049930) recently. Any idea what the best approach is to addressing these?

mgr0dzicki:
I believe this is because the new version of rustdoc format (v25) has just been released in nightly toolchain but it is not supported by `cargo-semver-checks` yet. @obi1kenobi, could you have a look at this? If I am right, the solution might be to fix the toolchain used by `cargo-semver-checks-action` to previous nightly until the new format of rustdoc is supported by `cargo-semver-checks`.

obi1kenobi:
If you're able to use stable Rust, I'd recommend using that for the GitHub Action to avoid exactly this kind of issue.

Integrating a new rustdoc format into cargo-semver-checks depends on a few upstream crates being released in sequence, and I happen to know that last week was an exam week for one of those crates' maintainers :) So we all do our best, but being volunteers we can't guarantee an SLA for supporting nightly.

Our own CI makes sure beta and stable Rust are supported, so I'd recommend using either beta or stable Rust on the action if at all possible.

zesterer:
Thanks!

Issue #403, opened 2023-04-22
Better documentaton of function parsers, pretty please


jamesyoungman:
The examples all (AFAICS) use `let` to define parsers.  But there are use-cases where this isn't convenient:

1. Unit testing - for more complex languages, it's helpful to expose parsers for parts of the language so that they can be unit-tested independently.
2. Beginners with chumsky - I'm trying to convert from nom to chumsky and the most obvious way to proceed is bottom-up; to implement chumsky parsers for the simplest parts of the grammar first, and then build upward.

Neither of these things seems very easy when constructing a correct impl Parser<...> declaration for the parser function's return type is entirely unobvious.   I mean, the parser itself seems straighforward (obviously I picked a trivial parser for the sake of the example):

    /// Accept either Unicode Dot Operator U+22C5 ("⋅") or Unicode Middle Dot U+00B7 ("·").
    fn maybe_normal_dot<'srcbody, I: Input<'srcbody>, E>() -> impl Parser<'srcbody, I, bool, E> {
        one_of("\u{22C5}\u{00B7}").to(true).or_not()
    }
    
but that parser can't be used anywhere (with chumsky 1.0.0-alpha.3) because the declared return type is missing a lot of trait bounds.   Some guidance from the examples or the documentation would be very welcome.

jamesyoungman:
I think the first step would be to better document the parameters of the Parser trait.  The documentation of the Input token type and the "extra" trait parameter seems to be very minimal.

Beyond that, an example parser in which many of the non-terminals are parsed by functions which have unit-tests.

If you want to be more ambitious, a tutorial showing the _incremental_ migration of a non-trivial parser to Chumsky from other frameworks (though my interest is primarily nom).

willothy:
I just migrated from nom and I've been using the `State` type / `with_state` functions to insert AST nodes into an arena, returning a key instead of the node itself. I would be open to contributing some documentation relating to this use case if that would be useful.

zesterer:
> I would be open to contributing some documentation relating to this use case if that would be useful.

That would be very much appreciated!

faassen:
I've been using a function-based approach with Chumsky 0.9, exactly because I wanted to make sure I could unit test. But this leads to some issues when recursion is involved, I think (I will explore more tomorrow).  And I now understand that the function signature will become more difficult in the Chumsky 1.0 world.

So let me rephrase the question: *how to unit test sub-parsers?* If we have an answer to that, we know what documentation to provide.

[...]

PR #412, opened 2023-05-04
Improve some documentation

### LINKED ISSUE #402: Parser::map_with_state example actually shows Parser::map_with_span
The doc comment for map_with_state looks like this (note that the example doesn't show map_with_state being used):

    /// Map the output of this parser to another value, making use of the parser's state when doing so.
    ///
    /// This is very useful for parsing non context-free grammars.
    ///
    /// The output type of this parser is `U`, the same as the function's output.
    ///
    /// # Examples
    ///
    /// ```
    /// # use chumsky::prelude::*;
    /// use std::ops::Range;
    ///
    /// // It's common for AST nodes to use a wrapper type that allows attaching span information to them
    /// #[derive(Debug, PartialEq)]
    /// pub struct Spanned<T>(T, SimpleSpan<usize>);
    ///
    /// let ident = text::ident::<_, _, extra::Err<Simple<char>>>()
    ///     .map_with_span(|ident, span| Spanned(ident, span))
    ///     .padded();
    ///
    /// assert_eq!(ident.parse("hello").into_result(), Ok(Spanned("hello", (0..5).into())));
    /// assert_eq!(ident.parse("       hello   ").into_result(), Ok(Spanned("hello", (7..12).into())));
    /// ```
    fn map_with_state<U, F: Fn(O, I::Span, &mut E::State) -> U>(
        self,
        f: F,
    ) -> MapWithState<Self, O, F>
    where
        Self: Sized,
    {
        MapWithState {
            parser: self,
            mapper: f,
            phantom: EmptyPhantom::new(),
        }
    }

CraftSpider:
Fixes #402
Steps towards #403, but not a full fix

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -863,7 +863,7 @@ pub trait Parser<'a, I: Input<'a>, O, E: ParserExtra<'a, I> = extra::Default>:
 
     /// Parse input as part of a token-tree - using an input generated from within the current
     /// input. In other words, this parser will attempt to create a *new* input stream from within
-    /// the one it is being run on, and the parser is was called on will be provided this *new* input.
+    /// the one it is being run on, and the parser it was called on will be provided this *new* input.
     /// By default, the original parser is expected to consume up to the end of the new stream. To
     /// allow only consuming part of the stream, use [`Parser::lazy`] to ignore trailing tokens.
     ///
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1828,6 +1828,76 @@ pub trait Parser<'a, I: Input<'a>, O, E: ParserExtra<'a, I> = extra::Default>:
     /// Boxing a parser is broadly equivalent to boxing other combinators via dynamic dispatch, such as [`Iterator`].
     ///
     /// The output type of this parser is `O`, the same as the original parser.
+    ///
+    /// # Examples
+    ///
+    /// When not using `boxed`, the following patterns are either impossible or very difficult to express:
+    ///
+    /// ```compile_fail
+    /// # use chumsky::prelude::*;
+    ///
+    /// pub trait Parseable: Sized {
+    ///     type Parser<'a>: Parser<'a, &'a str, Self>;
+    ///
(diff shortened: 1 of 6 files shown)


src/lib.rs
```diff
@@ -524,16 +535,61 @@ pub trait Parser<'a, I: Input<'a>, O, E: ParserExtra<'a, I> = extra::Default>:
...
+    /// #                     self.strings.push(str.to_string());
+    /// #                     Key(self.strings.len() - 1)
+    /// #                 }
+    /// #             }
+    /// #         }
+    /// #         pub fn get(&self, key: Key) -> &str {
+    /// #             &self.strings[key.0]
+    /// #         }
+    /// #     }
+    /// # }
```
zesterer, line 560: Good example! I'd even suddenly making the interner itself visible, or spending on `lasso`

src/lib.rs
```diff
@@ -806,7 +862,15 @@ pub trait Parser<'a, I: Input<'a>, O, E: ParserExtra<'a, I> = extra::Default>:
     }
 
     /// Parse input as part of a token-tree - using an input generated from within the current
-    /// input.
+    /// input. In other words, this parser will attempt to create a *new* input stream from within
+    /// the one it is being run on, and the parser is was called on will be provided this *new* input.
```
zesterer, line 866: is -> it

src/lib.rs
CraftSpider: By 'spending' do you mean pulling it in and actually using it? If you're fine with adding it as a dev dependency, I think that would be good - I'd be hesitant to expose the stub interner here - it's not the *worst* implementation but it's not one I'd advise exactly.

zesterer: Sorry, I think I'd had a bit to drink and was typing on mobile when I wrote this. Should have been:

> Good example! I'd even suggest making the interner itself visible or depending on `lasso` (i.e: using `lasso` is part of the doc example given that it's a commonly used interning crate)

zesterer:
Happy to merge?

CraftSpider:
Yep, ready when you are

PR #417, opened 2023-05-09
docs: add arena allocation examples

willothy:
Draft PR for now

### DIFF
--- a/src/extra.rs
+++ b/src/extra.rs
@@ -28,7 +28,7 @@ pub type Err<E> = Full<E, DefaultState, DefaultCtx>;
 
 /// Use specified state type, but default other types
 ///
-/// Use `State<S>` or `Full<E, S, C>` as the `E` type parameter of a parser to use a custom state type.
+/// Use `State<S>` or `Full<E, S, C>` as the `Extra` type parameter of a parser to use a custom state type.
 /// You can then use `parser().parse_with_state(&mut S)` to parse with a custom state
 ///
 /// # Examples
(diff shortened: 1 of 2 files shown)


zesterer:
This is looking great!

src/extra.rs
```diff
@@ -26,6 +27,21 @@ pub type Default = Full<DefaultErr, DefaultState, DefaultCtx>;
...
+///
+/// # Examples
+///
+/// One common use case for this is arena allocation of AST Nodes or tokens.
+///
+/// ```
+/// // Use a slotmap to allocate AST nodes, returning an id from each parser instead of a node
+/// pub fn parser<'src>() -> impl Parser<'src, BoxedStream<'src, Token>, NodeId, State<SlotMap<NodeId, ASTNode>> {
+///     // ...
+/// }
```
zesterer, line 42: I would probably avoid the use of `BoxedStream` here if only because it's likely to confuse users. Perhaps something like the following?

```rs
type ParserState = SlotMap<NodeId, ASTNode>;

pub fn parser<'src>() -> impl Parser<'src, &'src [Token], NodeId, extra::State<ParserState> {
    ...
}
```

src/extra.rs
```diff
@@ -26,6 +27,21 @@ pub type Default = Full<DefaultErr, DefaultState, DefaultCtx>;
 pub type Err<E> = Full<E, DefaultState, DefaultCtx>;
 
 /// Use specified state type, but default other types
+///
+/// Use `State<S>` or `Full<E, S, C>` as the `E` type parameter of a parser to use a custom state type.
```
zesterer, line 31: Do you mean `S` instead of `E`?

src/extra.rs
willothy: Sounds good! I'll update that.

src/extra.rs
willothy: I meant the Extra type parameter of Parser - I'll just say that instead to be more clear.

zesterer:
Looks good! Just a conflict to address and then I think this is ready for merge :)

willothy:
> Looks good! Just a conflict to address and then I think this is ready for merge :)

Awesome! Fixed the conflict, just a single line in extra.rs.

PR #437, opened 2023-06-05
Add support for unicode identifiers

### LINKED ISSUE #413: Unicode-aware text parsers
The current `text` parsers use ASCII-based handling entirely - this is fine for some use cases, but it would be nice for languages that want to actually handle unicode to be able to use the builtin `ident` and such parsers for it

TzviPM:
Fixes #413.

I gated this behind a feature called `unicode` so that it will be opt-in, including the dependency on `unicode-ident`. For now, I didn't enable the feature by default, so it won't be a breaking change.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -39,6 +39,8 @@ sync = ["spin"]
 
 # Adds impl of Parser for either::Either
 either = ["dep:either"]
+# Enable support for unicode identifiers
+unicode = ["dep:unicode-ident"]
 
 # An alias of all features that work with the stable compiler.
 # Do not use this feature, its removal is not considered a breaking change and its behaviour may change.
@@ -56,6 +58,7 @@ stacker = { version = "0.1", optional = true }
 regex = { version = "1.7", optional = true }
 spin = { version = "0.9", features = ["once"], default-features = false, optional = true }
 either = { version = "1.8.1", optional = true }
+unicode-ident = {version = "1.0.9", optional = true}
 
 [dev-dependencies]
 ariadne = "0.2"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -39,8 +39,6 @@ sync = ["spin"]
 
 # Adds impl of Parser for either::Either
 either = ["dep:either"]
-# Enable support for unicode identifiers
-unicode = ["dep:unicode-ident"]
 
 # An alias of all features that work with the stable compiler.
 # Do not use this feature, its removal is not considered a breaking change and its behaviour may change.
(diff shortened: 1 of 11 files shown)


zesterer:
> Thanks @zesterer. I love the work you've done on chumsky and ariadne and thrilled that I finally have a chance to contribute, however small.

Thanks! I really appreciate that :slightly_smiling_face: 

> I think that the current change of making ident behave with ascii or Unicode based on a flag is the right approach for those functions.

It's important to remember that cargo unifies feature flags across the dependency tree. This means that if one crate somewhere in the dependency tree gets a feature enabled, all other instances of that crate will similarly have it enabled. This means that the feature flag could cause subtle and incorrect 'spooky action at a distance' on the other side of the dependency tree and is why cargo guidance says that feature flags should only ever add API features rather than changing existing ones. For this reason, I don't think this behaviour should be controlled by a feature flag.

> I'm also open to keeping ident as-is and adding this as a new function: ident_Unicode with corresponding keyword_unicode

I think this is likely closer to the best result. I think my approach would be something like the following:

1) Move (most) of the existing text parsers into an `ascii` submodule

2) Create a `unicode` submodule with specialised versions of those in `ascii` that support unicode

```rs
pub mod text {
    pub mod ascii {
        pub fn ident(...) -> impl Parser { ... }
        pub fn keyword(...) -> impl Parser { ... }
    }
    pub mod unicode {
        pub fn ident(...) -> impl Parser { ... }
        pub fn keyword(...) -> impl Parser { ... }
    }
}
```

Although this would be a nominally breaking change (not really, `1.0` is still in alpha) I think this is a more rugged long-term approach. It also means that combinators like `newline` don't need to pay the extra cost of testing for unicode newlines for parsers that don't need it.

There's always the wider question posed by #237, that this PR slightly touches upon: just how exhaustive should the built-in combinators be when it comes to facilitating all of the different kinds of pattern users might want? For example, should `keyword`/`ident` also facilitate Swift-style emoji identifiers? I have no idea.

When I see a specific pattern being repeated *a lot* in user code, and in a consistent way that's easy to turn into a built-in combinator, then I think it's worth pulling that pattern up into a built-in combinator, but weighing those benefits against the maintenance cost and cognitive overload of having more combinators for users to sift through is a difficult calculus.

TzviPM:
I'm not sure which other parsers in `text` are ascii-specific, but as of this PR, I have moved the ones for `ident` and `keyword` into an `ascii` submodule and created a parallel `unicode` submodule with their unicode counterparts. I've also updated the tests to show how these might be used to support identifiers in other languages.

zesterer:
Thanks! I appreciate this was quite a bit of work, so I'm very grateful. Seems like there's just a minor conflict to address before merge.

zesterer:
Ah, looks like there are some failing tests (probably doc tests?) sorry :(

TzviPM:
Yup. I fixed ident but not keyword 🤦‍♂️

zesterer:
Thanks for the PR!

PR #439, opened 2023-06-07
Added with_slice, updated logos example

zesterer:


Amejonah1200:
It would be nice to have also a [lasso](https://crates.io/crates/lasso) integration.
This would allow making calls like `ident().intern()` possible.

examples/logos.rs
```diff
@@ -31,8 +33,6 @@ enum Token<'a> {
 
     #[regex(r"[ \t\f\n]+", logos::skip)]
     Whitespace,
```
Amejonah1200, line 35: Can be moved under the `#[derive(...)]` in `#[logos(skip r"[ \t\f\r]+")]`

Zij-IT:
> It would be nice to have also a [lasso](https://crates.io/crates/lasso) integration. This would allow making calls like `ident().intern()` possible.

Would this not be equivalent to:

```rust
ident().map_with_state(|i, _, rodeo: Rodeo| rodeo.get_or_intern(i))
```

Amejonah1200:
> > It would be nice to have also a [lasso](https://crates.io/crates/lasso) integration. This would allow making calls like `ident().intern()` possible.
> 
> Would this not be equivalent to:
> 
> ```rust
> ident().map_with_state(|i, _, rodeo: &Rodeo| rodeo.get_or_intern(i))
> ```

Yes, I will use trait extensions for that then.

PR #442, opened 2023-06-08
Chumsky: Getting a Book

Zij-IT:
Heyyo,

I figured I would take the leap and go ahead and try and cobble together the various resources for Chumsky into a book using [mdbook](https://rust-lang.github.io/mdBook/index.html).

If there are pieces that are missing, or that you would like included in the book, please let me know! There is also probably a typo hiding somewhere in the book. It currently aligns with version 1.0.0-alpha.4 of Chumsky. Hopefully when it comes time for the official 1.0.0 release, it will just be a simple replacement of `1.0.0-alpha.4` with `latest`, but even if not, I'll fix the broken links.

I tried to pay special attention to the links to make sure that every one of them works and sends to the right place. If one is broken, please let me know and I will get to it when I can.

Thanks for your time!


Zij-IT:
So, I am not exactly sure what you would prefer to have happen with the `guide`, but given that the book ate it, I figured it was correct to remove `guide.rs`, and allow the tests to pass.

zesterer:
Thanks for the PR! I would say that I'm opposed to moving the guide in general. I considered mdbook prior to using creating it, but decided against it in favour of hosting on docs.rs since it has several distinct advantages:

- Examples can be checked with CI easily as regular doc tests
- The guide can link directly to items in the crate itself, making navigation simpler
- Documentation on crate items can link directly to relevant sections in the guide
- Simpler CI: no need to host everything externally, have a dedicated static site, etc.
- The guide automatically supports versioning and is tied to the implementation of the crate that it references

intro/src/foo/numbers.md
```diff
@@ -0,0 +1,112 @@
...
+
+The answer is that Chumsky's parsers are *lazy*: they will consume all of the input that they can and then stop.
+If there's any trailing input, it'll be ignored.
+
+This is obviously not always desirable. If the user places random nonsense at the end of the file, we want to be
+able to generate an error about it! Worse still, that 'nonsense' could be input the user intended to be part of
+the program, but that contained a syntax error and so was not properly parsed. How can we force the parser to consume
+all of the input?
+
+To do this, we can make use of two new parsers: the `then_ignore` combinator and the `end` primitive.
```
zesterer, line 61: Note that this is no longer true: chumsky parsers are now forced to consume everything by default (as if they had an implicit `.then_ignore(end())`). The old behaviour can be recovered with the `.lazy()` combinator. I think this section can be removed from the tutorial entirely.

intro/src/chumsky.md
```diff
@@ -0,0 +1,38 @@
...
+
+```
+test chumsky ... bench:   4,782,390 ns/iter (+/- 997,208)
+test pom     ... bench:  12,793,490 ns/iter (+/- 1,954,583)
+```
+
+This is obviously slower than a handwritten parser, but that's okay. Chumsky's
+goal is not to be the fastest, and if you're far enough in development that
+parsing performance is the actual problem, it's likely time to be moving to a
+hand-written parser anyway.
```
zesterer, line 38: This old README information is out of date, I'm currently in the process of writing a new one.

Zij-IT:
Sorry it took so long to get back to this. Anyhow, I can definitely understand the wanting to stay by the guide as written, so I will go ahead and close this.

Issue #431, opened 2023-05-22
Implement Parser for either::Either


zesterer:
Very useful for building up parsers at runtime without `.boxed()`.

Zij-IT:
So here's the question: Should this be handled by `chumsky`, or should a pull request be made for `either`, which contains the implementation?

I only know from `serde` that it provides the traits and has the other libraries have the impls instead of trying to maintain those itself.

zesterer:
I would say that this is *probably* something to put behind an `either` feature flag in chumsky. That said, I think `itertools` also has its own implementation of `Either` :upside_down_face: so we can't please everybody. The former is pretty common, however.

PR #443, opened 2023-06-11
either: add `Parser` impl for `either::Either`

Zij-IT:
Just adds a feature-gated `Parser` impl for `Either`. Otherwise nothing special about it.

Addresses #431


zesterer:
Thanks! This is ideal :)

PR #451, opened 2023-06-18
Fix JSON example

OlofBlomqvist:
The current json example produces incorrect error reports after objects and arrays, and in some cases causes valid objects to not be included in the result. This PR should fix that, altough, tbh I only just started looking at chumsky today so this might not be the best way to solve it(?)

Examples of json that produces incorrect errors prior to this change:

```json
{
    "A": {
        "B": 123
    },
    "C": {
        "D": 456
    },
    "E": [
        "F"
    ],
    "G": "done"
}
```

```json
{
    "A": {
        "B": "123",
        "C": [1,2,3],
        "D": 123
    }    
}
```

bild
bild


zesterer:
Thanks so much for noticing and fixing this!

PR #452, opened 2023-06-19
feat: re-implement semantic indent parser

willothy:
I needed the semantic indent parser for a project, but it seems to have been removed - this is my re-implementation of it. Heavily based on the version from `0.9.x`.

Draft PR as this may not be an ideal implementation, and I'm not sure what module it should be added to (created `indent.rs` for now).

### DIFF
--- a/src/indent.rs
+++ b/src/indent.rs
@@ -21,7 +21,7 @@ where
     T: Parser<'a, I, O, E> + Clone,
     F: Fn(Vec<O>, S) -> O + Clone,
 {
-    fn collapse<'a, O, S>(
+    fn collapse<O, S>(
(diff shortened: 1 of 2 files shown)


fowenix:
The [indent example](https://github.com/zesterer/chumsky/blob/main/examples/indent.rs) showcases semantic indentation for the current version of chumsky, not sure if this is really needed

willothy:
> The [indent example](https://github.com/zesterer/chumsky/blob/main/examples/indent.rs) showcases semantic indentation for the current version of chumsky, not sure if this is really needed

Didn't realize that! My bad

willothy:
> Yep, the new context-sensitive parsing combinators are much more general and powerful, allowing the parsing of many kinds of context-sensitive syntax.

Cool, I'll close this then. Thanks!

PR #453, opened 2023-06-20
Add map_tuple nightly combinator

fowenix:


zesterer:
The implementation looks good to me. I wonder, though, whether `map_tuple` is a good name. After all, the combinator has the effect of *hiding* the tuple such that the user shouldn't even need to think about it. Perhaps something like `map_group` is a better name?

PR #454, opened 2023-06-20
Rename context combinators

fowenix:


bew:
Why?
Was it discusser previously?

zesterer:
Hmmm, what value does this rename bring?

fowenix:
I felt like `ignore_with_ctx` and `then_with_ctx` weren't the best names for what they do.
But after some more thought, I now think `then_with_ctx` is fine but `ignore_with_ctx` feels off

fowenix:
Maybe renaming `ignore_with_ctx` into `then_into_ctx` where this shows that the ctx is being consumed, while `then_with_ctx` shows it is being group by `with`

zesterer:
Oh I didn't mean that as an instruction, just as an opening point of discussion. I'm not convinced by the names I suggested myself.

fowenix:
> a.ignore_then_ctx(b)
a.then_ctx(b)

IMO this is the most convincing and asking ChatGPT to generate some alternatives names didn't give any viable ones.

zesterer:
Well a problem with `.then_ctx` is that it implies that the second part is the context, which it isn't. The 'most correct' names would probably be `.ctx_then` and `.ctx_ignore_then`, and are potentially a little easier to quickly spot as context-sensitive combinators.

Regardless, I think this is something better left to an issue with discussion for the time being.

PR #457, opened 2023-06-25
Add SimpleSpan::splat

dyslexicsteak:
SimpleSpan::splat creates a SimpleSpan from a single offset using it as both the start and end. This is more convenient for making an EOI span than passing the same argument twice into SimpleSpan::new.

### DIFF
--- a/src/span.rs
+++ b/src/span.rs
@@ -66,6 +66,11 @@ impl<T> SimpleSpan<T> {
         }
     }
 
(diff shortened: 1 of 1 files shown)


dyslexicsteak:
It does seem that there might be a problem with splat, because it'll require Clone.

dyslexicsteak:
I'll fix the commits and reopen a PR, sorry for spontaneity.

PR #458, opened 2023-06-25
Add SimpleSpan::splat

dyslexicsteak:
SimpleSpan::splat creates a SimpleSpan from a single offset using it as both the start and end. Only works for types which are Clone. Meant to be more convenient for creating an EOI span than duplicating the argument for SimpleSpan::new.


zesterer:
Thanks!

zesterer:
Looks like it needs a `cargo fmt`.

dyslexicsteak:
> Looks like it needs a `cargo fmt`.

I had ran one, seems I invalidated it after. It added a trailing comma and did some whitespace fixes.

Issue #444, opened 2023-06-15
Normalize spelling of memoization/memoisation


bglw:
Currently the crate feature is `memoization`, but the function & type are `memoised` / `Memoised`. I don't have much of a preference either way — but it would be good to have them match 🙂

https://github.com/zesterer/chumsky/blob/da41f156f45d12c27c94e7c93e18ea33514a0ed6/src/lib.rs#L735-L736

zesterer:
Ah, the pains of being a British developer on a US-oriented internet! Seems like some typos slipped through. I'll put this on my todo list.

bglw:
I hear ya from New Zealand — not a typo to me 😎

Zij-IT:
As one of those US people, I am surprised I didn't catch the discrepancy, because "memoised" is definitely wrong ;)

Zij-IT:
Closable :)

PR #461, opened 2023-06-27
memo-name: correct inconsistent spelling

Zij-IT:
Addresses #444 by moving everything to the American-English 'z' instead of 's'.

I suppose this technically is a breaking change, because `Memoise` was public, but eh.


zesterer:
Thanks!

Issue #398, opened 2023-04-11
Suggestion for a text parser: `decimal_number()`


LikeLakers2:
Hi! I was looking at `chumsky` and found that there doesn't appear to be a built-in parser for recognizing decimal numbers, such as `4.21` or `4.2e-7`. So I'd like to suggest that one be added, since understanding number formats may be a common task for many chumsky parsers.

Looking through the examples, it appears the JSON parser example already implements a number parser:

https://github.com/zesterer/chumsky/blob/f664d7d7f36205bca58842233431927c14be69c0/examples/json.rs#L22-L37

so you could feasibly base the built-in parser off that code, instead of needing to figure it out from scratch. (Though I would recommend having the built-in parser return a string/vec rather than an f64, just in case the decimal number is more precise than an f64 can handle.)

zesterer:
This is something I'd definitely like to integrate at some point, but it feels like this is better suited to an extension crate (or at least being behind a feature flag).

There are also a lot of subtle complications since number parsing varies a lot between languages. Even between C++ and Rust, two C-likes, there's the difference between `'` and `_` separators used to visually split up numbers.  This gets even more subtle when you get into scientific notation and the question of how to actually take the parsed number and precisely and correctly convert it into an actual float with the semantics that the language expects.

I'm happy to accept a PR for a simple form of this though, something that accepts an `int`, a full stop, and any number of trailing digits: I think that's a simple enough common denominator to cover a lot of small langs without introducing confusing functionality.

LikeLakers2:
> This gets even more subtle when you get into scientific notation and the question of how to actually take the parsed number and precisely and correctly convert it into an actual float with the semantics that the language expects.

So wait, are you suggesting that this parser return a f32/f64 with the input it matched? Or do you think it should return a string/Vec<u8> with the input it matched?

zesterer:
I'm not sure, to be honest! Outputting `f32`/`f64` means that we'd need to think carefully about how it deals with precision issues. Outputting the slice matched by it would just shift that responsibility on to the user.

LikeLakers2:
Then might I suggest having more than one function, like `nom` does? To explain the parsers that `nom` has:

* Two parsers that parse the stream for a number, and return either [a `f32`](https://docs.rs/nom/7.1.3/nom/number/complete/fn.float.html) or [a `f64`](https://docs.rs/nom/7.1.3/nom/number/complete/fn.double.html).
[...]

PR #462, opened 2023-06-27
[WIP] Add parsing of lexical numbers

Zij-IT:
I figured I would submit a WIP pull request to get your opinion on it, before I spend time writing documentation (as well all know that takes forever), and make any changes that you feel are necessary.

This pull request adds the following public elements:
* `Number` - Parses a number given a specific format provided by the [`lexical`](https://docs.rs/lexical/latest/lexical/) crate.
* `number` - Which is just the constructor function.
* `format` - A reexport of [`lexical::format`](https://docs.rs/lexical/latest/lexical/format/index.html)

This pull request changes the the following:
* Relaxes the bound on `InputRef::skip_bytes` to allow for `number` to work with `SliceInput`. I am admittedly not sure if this is the right call, but it seemed like the previous bound was unnecessarily strict.

Questions:
* Does this look like the API that you would be expecting for a parser like this?
* Are there any surprises that I should be aware of that force `Number` to require `I: StrInput`?
* Is there any reason for `skip_bytes` to require `StrInput` that I should be aware of?
* Should the aforementioned bound be set-back to `StrInput`?.


Zij-IT:
Ah forgot to tag the issue. #398

src/input.rs
```diff
@@ -1136,12 +1136,12 @@ impl<'a, 'parse, I: Input<'a>, E: ParserExtra<'a, I>> InputRef<'a, 'parse, I, E>
         unsafe { self.input.span(before.offset..self.offset) }
     }
 
-    #[cfg(feature = "regex")]
     #[inline(always)]
-    pub(crate) fn skip_bytes<C>(&mut self, skip: usize)
+    #[cfg(feature = "regex")]
+    #[cfg(feature = "lexical-numbers")]
```
zesterer, line 1141: I think `cfg(any(feature = "regex", feature = "lexical-numbers"))` is the proper thing to do here.

src/number.rs
```diff
@@ -0,0 +1,165 @@
...
+    use lexical::format::RUST_LITERAL;
+
+    // These have been shamelessly yanked from the rust test-float-parse suite.
+    // More specifically:
+    //
+    // https://github.com/rust-lang/rust/tree/64185f205dcbd8db255ad6674e43c63423f2369a/src/etc/test-float-parse
+    mod rust {
+        use super::*;
+
+        const FLOAT: Number<RUST_LITERAL, &str, extra::Default, f64> = number();
```
zesterer, line 66: It's a bit of a shame that we end up needing to specify the type parameters like `number::<RUST_LITERAL, _, _, _>()` each time due to the const generic, but by the looks of things there might not be a way around this.

zesterer (approved):
This seems reasonable. That said, it might be worth putting behind an `unstable` feature flag too for now (with `lexical-numbers = ["lexical", "unstable"]`). There's enough weirdness with the design (not your fault, just a product of `lexical`'s API) that it would be nice to change this in breaking ways post-release.

src/number.rs
Zij-IT: I agree, but I have no clue how we could even try to work around this, because `parse_partial` uses the `RUST_LITERAL` generically.

Unrelated, but It may also be worth it to use `const_assert` to make sure that the literal provided is [valid](https://docs.rs/lexical/latest/lexical/fn.format_is_valid.html).

Zij-IT:
@zesterer Should be merge-ready :)

Issue #145, opened 2022-05-19
Performance of right-associative operators


01mf02:
Hi,

let me first thank you for your excellent work on `chumsky`, `ariadne` and other crates such as `flume` that I had the pleasure to use.

I am using `chumsky` in my [`jaq`](https://github.com/01mf02/jaq) crate. Somebody has hit a performance bottleneck in my parser, see 01mf02/jaq#7.
I was able to trace down the problem to my implementation of right-associative operators, using `Parser::foldr`.
I am basically following your technique in [`nano_rust.rs`](https://github.com/zesterer/chumsky/blob/master/examples/nano_rust.rs) to implement operators with differing precedence, but I also require right-associative operators.
A simplified version that demonstrates the problem is shown below:

~~~ rust
use chumsky::{error::Simple, prelude::*};

fn bin<P, O>(prev: P, op: O) -> impl Parser<char, usize, Error = P::Error> + Clone
where
    P: Parser<char, usize> + Clone,
    O: Parser<char, char, Error = P::Error> + Clone,
{
    // right-associative: extremely slow!
    /*
    let args = prev.clone().then(op).repeated().then(prev);
    args.foldr(|(a, _op), b| a + b)
    */

    // left-associative
    let args = prev.clone().then(op.then(prev).repeated());
    args.foldl(|a, (_op, b)| a + b)
}

fn main() {
    let mut expr = Recursive::declare();

    let int = text::int::<char, Simple<char>>(10)
        .from_str()
        .unwrapped()
        .or(expr.clone().delimited_by(just('('), just(')')));
    let add = bin(int, just('+')).boxed();
    let sub = bin(add, just('-')).boxed();
    let mul = bin(sub, just('*')).boxed();
    let div = bin(mul, just('/')).boxed();

    expr.define(div);

    assert_eq!(expr.parse("3"), Ok(3));
    assert_eq!(expr.parse("((((1+2-3*4/5))))"), Ok(15));
}
~~~

When using the right-associative variant of the `bin` function, then the parser gets quite a bit slower; in particular when adding more parentheses in the parsed expression.
Do you have any idea how to make right-associative operators faster?

zesterer:
Apologies about not responding earlier. This behaviour is rather strange. I'll try to get the time to do some local testing.

zesterer:
Hopefully #288 will go a long way toward improving the writing of operator parsing.

01mf02:
[...]

Issue #51, opened 2021-12-09
Add a Pratt parsing combinator


zesterer:

kevinbarabash:
Would this allow a person to embedded a Pratt parser for algebraic expressions within in a parent parser for a programming language?

zesterer:
Yes, that would be the goal.

kevinbarabash:
That sounds awesome.  I'm curious if you had any thoughts on what the API might look like.

I wrote a Pratt [parser factory](https://github.com/math-blocks/math-blocks/blob/master/packages/parser/src/factory.ts) in TypeScript a while ago.  It's surprisingly short, but it also doesn't support post-fix operators and there's some hardcoded logic in it, e.g. checking for unmatched parens.

I'm still relatively new to rust, but would be interested in helping out with this issue.

zesterer:
I've not really had too many thoughts about it yet, to be honest. Perhaps I'll try some experimenting next week.

zesterer:
`zero-copy` is getting a pratt parser combinator. See #288.

Zij-IT:
Closable with #464

PR #464, opened 2023-06-28
The final pratt parser PR?

Zij-IT:
Alright, so after some more thought and deciding I was going to sit down and take a final crack at it, this is what I was able to come up with that best resembles the API that was discussed in both #380, #288 and #144.

The example is now much slimmer, with it being as follows:

```rust
enum Expr {
    Literal(i64),
    Add(Box<Expr>, Box<Expr>),
    Sub(Box<Expr>, Box<Expr>),
    Mul(Box<Expr>, Box<Expr>),
    Div(Box<Expr>, Box<Expr>),
}

impl std::fmt::Display for Expr {
    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
        match self {
            Self::Literal(literal) => write!(f, "{literal}"),
            Self::Add(left, right) => write!(f, "({left} + {right})"),
            Self::Sub(left, right) => write!(f, "({left} - {right})"),
            Self::Mul(left, right) => write!(f, "({left} * {right})"),
            Self::Div(left, right) => write!(f, "({left} / {right})"),
        }
    }
}

let atom = text::int::<_, _, extra::Default>(10)
    .from_str()
    .unwrapped()
    .map(Expr::Literal);

let operator = choice((
    PrattOp::new_left(just('+'), 0, |l, r| Expr::Add(Box::new(l), Box::new(r))),
    PrattOp::new_left(just('-'), 0, |l, r| Expr::Sub(Box::new(l), Box::new(r))),
    PrattOp::new_right(just('*'), 1, |l, r| Expr::Mul(Box::new(l), Box::new(r))),
    PrattOp::new_right(just('/'), 1, |l, r| Expr::Div(Box::new(l), Box::new(r))),
));

let expr = atom.pratt(operator.padded_by(just(' ')));
```

The most notable inclusion into this is `PrattOp`, and the privatization of `Assoc`, `InfixPrecedence` and `Strength`. Of note: This iteration defines `Strength` as a simple `u8`, as it was easier to work with, though `PrattOp` could be made to be generic over the type of `Strength` as it was before.

How do you like this version of the pratt API?


zesterer:
This is a really nice API! Much, much closer to the sort of design I was hoping for, thanks so much for pushing this forward :smiley:

There are a few minor things I think could be improved, but they're relatively trivial:

- `PrattOp::new_left` seems a bit wordy, I think `pratt_left` (and the same for its right-associative brother) would be a little more terse
- `new_left`/`new_right` only accepting a function pointer is a little awkward. Given that these are used in a `choice`, I don't there's any specific reason why these couldn't accept an `impl Fn`, but perhaps there's something here that I'm not seeing (perhaps because the output types of the `choice` need to unify into `PrattOpOutput`?).

The last of these is the one that I think concerns me the most. Using the `Fn` traits is one of the real strengths of chumsky (both in terms of API flexibility and the compiler's ability to optimise the hell out of everything) and this somewhat flies in the face of that. If it makes the implementation more feasible, I think removing the `choice` and shoving them all into a tuple parameter would be worth doing, even if it does make dealing with padding and the like a little more awkward:

```rs
let op = |c| just(c).padded_by(just(' '));

let expr = atom.pratt((
    PrattOp::new_left(op('+'), 0, |l, r| Expr::Add(Box::new(l), Box::new(r))),
    PrattOp::new_left(op('-'), 0, |l, r| Expr::Sub(Box::new(l), Box::new(r))),
    PrattOp::new_right(op('*'), 1, |l, r| Expr::Mul(Box::new(l), Box::new(r))),
    PrattOp::new_right(op('/'), 1, |l, r| Expr::Div(Box::new(l), Box::new(r))),
));
```

Zij-IT:
## Shorthand Functions

Yeah I could totally see that. I went with `(left|right)_infix` following the idea behind the `text` constructors, so that in the future we will have the freedom to use `(left|right)_(pre|post)fix`.

## The Fn Family

> new_left/new_right only accepting a function pointer is a little awkward.

I will admit it is pretty awkward considering the rest of the crate uses of the `Fn` family. A minimized version of the issue that I am having would be:

```rust
fn test<I, Ex, E>() -> impl Parser<I, Fn(Ex, Ex) -> Ex, E> {
    choice((
        just('+').to(|a, b| a),
        just('-').to(|a, b| a),
        just('*').to(|a, b| a),
        just('/').to(|a, b| a),
    ))
}
```

This doesn't compile, because Rust can't know which closure is going to be returned, so we have to use indirection, or keep it simple with a function pointer. I gave it many attempts to use `&dyn Fn(Expr, Expr) -> Expr` to avoid an allocation with `Box`, but I can't get the parser to return a value that is borrowed from itself, without Rust throwing a fit about the lifetimes. Another alternative would be to use `Box<dyn (Fn(Expr, Expr) -> Expr) + Clone>` and require `Clone` for the closures. Though that would add many an allocations to the parser itself, because any time it successfully parses an operator, it would have to `Box` the result.

If someone is willing to bite the bullet and give it a shot without using function pointers I am totally for it, but this is about the fourth time I have hit the wall while trying to use the `Fn` trait family.

## Pre- and Postfix Operators

I actually have given these a fair bit of thought for the `Pratt` parser. I would love to be able to do something like:

```rust
pub struct Pratt<Atom, InfixOps, Expr, Op, Ex, PrefixOps = (), PostfixOps = ()> {
    pub(crate) atom: Atom,
    pub(crate) infix_ops: InfixOps,
    pub(crate) prefix_ops: PrefixOps,
    pub(crate) postfix_ops: PostfixOps,
    pub(crate) phantom: PhantomData<(Expr, Op, Ex)>,
}
```

And then have different `pratt_parse` depending on if the `PrefixOps` and/or `PostfixOps` are non-`()`, have optimized behavior:

```rust
// If both ()
let mut left = self.atom.go::<M>(inp)?;
loop {
    let pre_op = inp.save();
    let (op, prec) = match self.infix_ops.go::<Emit>(inp) {
        Ok(PrattOpOutput(prec, build)) => {
            if prec.strength_left().is_lt(&min_strength) {
                inp.rewind(pre_op);
                return Ok(left);
            }
            (build, prec)
        }
        Err(_) => {
            inp.rewind(pre_op);
            return Ok(left);
        }
    };

    let right = self.pratt_parse::<M, _>(inp, Some(prec.strength_right()))?;
    left = M::combine(left, right, op);
}

// If postfix != ()
let mut left = self.atom.go::<M>(inp)?;
loop {
    // Before parsing the righthand side, attempt postfix op
    let pre_op = inp.save();
    match self.postfix_ops.go::<Emit>(inp) {
        Ok(PrattOpOutput(prec, build)) => {
            if prec.strength_left().is_lt(&min_strength) {
                inp.rewind(pre_op);
                return Ok(left);
            }

            let right = self.pratt_parse::<M, _>(inp, None)?;
            left = M::combine(left, right, build);
        }
        Err(_) => {
            inp.rewind(pre_op);
        }
    }

    /* Parse RHS like above*/
}

// If prefix != ()
// Attempt to parse a prefix operator before parsing the LHS
let pre_op = inp.save();
let mut left = match self.prefix_ops.go::<Emit>(inp) {
    Ok(PrattOpOutput(prec, build)) => {
        let right = self.pratt_parse::<M, _>(inp, prec.stength_right())?;
        M::map(right, build)
    }
    Err(_) => {
        inp.rewind(pre_op);
        self.atom.go::<M>(inp)?
    }
};

/* Parse ops and right like before */
```

I don't know if Rust will let me do that without specialization. The way around that would be to have have various ones that represent the combinations `PrattWithPrefix`, `PrattWithPostfix`, `PrattWithAll` and `PrattOnlyInfix`. If we allow the infix-operators to be optional, we will need in total 8 combinators :D

## Finishing Words

I hope this wasn't too long, but I wanted to make sure I provided a thorough answer as to why it's not what was initially hoped, and how it can be extended in the future to allow for both prefix and postfix operators.

zesterer:
Thanks for the response :) I'm going to have a crack and making this work with `impl Fn` tomorrow, I think I have some ideas. I'll let you know if any of them work out!

Zij-IT:
I present you a fully-featured Pratt parser! Infix operators must be present, but `Prefix` and `Postfix` are optional, and each of them use their optimized parsing path thanks to the power of generics!

Adds:
* `Pratt::with_prefix_ops` - Used to add prefix operators to the pratt parser
* `Pratt::with_postfix_ops` - Used to add postfix operators to the pratt parser
* `prefix` - Used to construct a `PrefixOp` operator. Shorthand for `PrefixOp::new`
* `postfix` - Used to construct a `PostfixOp` operator. Shorthand for `PostfixOp::new`

There are a couple things that I would like to be improved, but that I don't feel are blockers:
* `Pratt::with_prefix_ops` and `Pratt::with_postfix_ops` currently are too loose with the generic for the passed in parser, which results in very complex errors when the type isn't the right kind of parser.
* `Pratt::with_prefix_ops` and `Pratt::with_postfix_ops` should only accept tuples of their respective types. Currently you can accidentally foot-gun yourself by using `postfix` as a parser for `with_prefix_ops`.

Once you are able to get `impl Fn` working, I will port it over to this :)

zesterer:
Oh wow! That's very impressive. Amazing work.

For the sake of not holding this up any longer, I think I'd propose to merge this but behind the `unstable` flag for the time being. What do you think?

Zij-IT:
@zesterer Gladly! And when it's merged I believe the following can be closed:
* #380
* #144
* #288 
* #145
* #51

stefnotch:
@Zij-IT I believe this doesn't get said often enough, so I'll say it: *Thank you* so much for donating your valuable time to create this excellent pull request!
I'm definitely looking forward to being able to use Chumsky in my math parsing projects.

PR #465, opened 2023-06-28
Fixes for lexical-numbers

Zij-IT:
Hey, this is just a quick PR to implement a QoL fix, and an implementation fix:
* `Number` should always be copyable
* reexport both `number` function and `number` module

I also took the time to update the example for showcasing in the related issue #398

### DIFF
--- a/examples/json.rs
+++ b/examples/json.rs
@@ -19,22 +19,28 @@ enum Json {
 
 fn parser<'a>() -> impl Parser<'a, &'a str, Json, extra::Err<Rich<'a, char>>> {
     recursive(|value| {
-        let digits = text::digits(10).slice();
+        #[cfg(not(feature = "lexical-numbers"))]
+        let number = {
+            let digits = text::digits(10).slice();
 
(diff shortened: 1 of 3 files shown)


zesterer:
I think I'm going to reject this PR for two reasons:

1) I accidentally just made the same change to the `Copy`/`Clone` impls for `Number` in another commit before seeing this PR.

2) Swapping out the numeric parser in the benchmark *feels* like cheating: there's no reason why you couldn't also give `nom`/`pom`/`winnow` the advantage of plugging into `lexical`, and I don't think that doing so for chumsky specifically really tells us anything particularly insightful, other than that rich error generation is expensive (which we already know from the rich equivalents of the benchmarks).

Zij-IT:
Totally get it! I have to remember while working with chumsky that if I am using `derive` I am probably doing something wrong.

zesterer:
Yeah, it's a bit of an annoyance. Hopefully one day derivation of traits for types that don't mention their type parameters per-value will be fixed in rustc. We can only hope...

PR #466, opened 2023-06-29
Add IoInput for using seekable readers as a parse input

CraftSpider:
Just what it says on the tin. Not sure how performant this is, but it uses a BufReader to avoid being *too* bad for the common cases.


zesterer (approved):
Seems sensible to me!

PR #467, opened 2023-06-29
Update intro.md

Andrew-Hanson:
updated links in intro.md


zesterer:
This change doesn't appear to work. What's the intention here?

Andrew-Hanson:
> This change doesn't appear to work. What's the intention here?

Sorry. I tried following one of the links and it did not work, and I saw the links included a _num_ prefix. The files in the repository did not include this in their name so I changed the link to reflect this. However, this change did not create working links.

Is there a page associated with these files other than the repository page, like a GitHub pages site?

If not, I have updates the links again to point to the .md documents in the repository, and I actually tested them this time. Sorry about that.

zesterer:
> Is there a page associated with these files other than the repository page, like a GitHub pages site?

Yep, check out [the guide](https://docs.rs/chumsky/1.0.0-alpha.4/chumsky/guide/index.html)

Andrew-Hanson:
I see. So the links are correct when used on the site. They’re relative links instead of absolute links, so they don’t work when you click on them while in the git repository.

zesterer:
Yes, unfortunately. I may add a `README.md` in the `guide/` directory that points readers to the docs.

PR #468, opened 2023-07-02
Add JustAnyCase primitive

lumenian:
Hello,

This is my first contribution to Chumsky, so any suggestions for possible improvements are welcome.

I've noticed there is no section for text-only primitives in `meet_the_parsers.md`. Should I add one?


zesterer:
Is this parser akin to `keyword`? What should happen if the parser `just_any_case("func")` sees the input `"function"`? Does it parse the first four characters and leave the remainder trailing, or does it operate like `keyword` and reject the whole thing?

lumenian:
> Is this parser akin to `keyword`? What should happen if the parser `just_any_case("func")` sees the input `"function"`? Does it parse the first four characters and leave the remainder trailing, or does it operate like `keyword` and reject the whole thing?

It's intended to behave identically to the `Just` primitive, except for matching in a case-insensitive way. So the following produces two identical errors:
```rust
let first = just::<_, _, extra::Err<Rich<char>>>("func")
    .parse("function")
    .into_errors();
dbg!(first);

let second = just_any_case::<_, _, _, extra::Err<Rich<char>>>("func")
    .parse("function")
    .into_errors();
dbg!(second);
```
```
first = [
    found ''t'' at 4..5 expected something else,
]
second = [
    found ''t'' at 4..5 expected something else,
]
```

zesterer:
Looks like it'll need a rebase before merge, sorry!

PR #469, opened 2023-07-02
pratt: fix and add clones

Zij-IT:
This just adds some quick fixing to `main` that was missed in the original PR. I missed some `Clone` impls and had forgotten to move some away from `derive`.

I only noticed when I went to use it in my own project, and so here we are :D


src/pratt/ops.rs
```diff
@@ -26,6 +26,62 @@ pub struct InfixPrefixPostfix<P1, P1O, P2, P2O, P3, P3O> {
     pub(crate) phantom: PhantomData<(P1O, P2O, P3O)>,
```
zesterer, line 26: Should these be `EmptyPhantom` instead of `PhantomData`?

Zij-IT: Out of curiosity, is `EmptyPhantom` made so that the following are equivalent?

```rust
// So that this:
#[derive(Clone)]
pub struct InfixPrefix<P1, P1O, P2, P2O> {
    pub(crate) infix: P1,
    pub(crate) prefix: P2,
    pub(crate) phantom: crate::EmptyPhantom<(P1O, P2O)>,
}

// is the same as this:
impl<P1, P1O, P2, P2O> Clone for InfixPrefix<P1, P1O, P2, P2O>
where
    P1: Clone,
    P2: Clone,
{
    fn clone(&self) -> Self {
        Self {
            infix: self.infix.clone(),
            prefix: self.prefix.clone(),
            phantom: EmptyPhantom::new(),
        }
    }
}
```

I ask because when I attempt to run my [code](https://github.com/Zij-IT/lamb/tree/feature/error-msgs) after swapping the hand-written `Clone` for `EmptyPhantom` + `derive` then I get an error that `PrattOpOutput<fn(Expr) -> Expr>` must implement `Clone` as seen here:

<details>
<summary>Compiler Error</summary>
<pre>
error[E0277]: the trait bound `pratt::ops::PrattOpOutput<fn(Expr, Expr) -> Expr>: Clone` is not satisfied
   --> src/parse.rs:62:20
    |
62  |         let expr = recursive(|expr| {
    |                    ^^^^^^^^^ the trait `Clone` is not implemented for `pratt::ops::PrattOpOutput<fn(Expr, Expr) -> Expr>`
    |
    = note: required for `InfixPrefix<Choice<...>, ..., ..., ...>` to implement `Clone`
    = note: the full type name has been written to 'lamb/lamb-rs/target/release/deps/lamb-d2f0d7d3cde0c777.long-type-9871736043221929107.txt'
    = note: 1 redundant requirement hidden
    = note: required for `Pratt<_, Expr, Full<..., ..., ...>, ..., ...>` to implement `Clone`
    = note: the full type name has been written to 'lamb/lamb-rs/target/release/deps/lamb-d2f0d7d3cde0c777.long-type-18055386955771684395.txt'
note: required by a bound in `recursive`
   --> chumsky/src/recursive.rs:271:30
    |
267 | pub fn recursive<'a, 'b, I, O, E, A, F>(f: F) -> Recursive<Direct<'a, 'b, I, O, E>>
    |        --------- required by a bound in this function
...
271 |     A: Parser<'a, I, O, E> + Clone + MaybeSync + 'b,
    |                              ^^^^^ required by this bound in `recursive`
</pre>
</details>

When using the hand-written implementation however I get no compiler error. Here is the [patch](https://github.com/zesterer/chumsky/files/11939868/patch.txt) with the difference.

zesterer: > Out of curiosity, is `EmptyPhantom` made so that the following are equivalent?

No. The issue with deriving clones is in the compiler-provided macro itself unfortunately. However, `EmptyPhantom` *does* mean that the compiler don't erroneously infer `!Send`/`!Sync`/etc. traits for the type.

Zij-IT: Ah, I see. I have added a commit which replaces all uses of `PhantomData` added by the `pratt` PR  with `EmptyPhantom`

Zij-IT:
Um... should I go ahead and fix the unrelated clippy errors?

zesterer:
Argh, that's a bit annoying. I'll try to fix these in the next day or two.

zesterer:
Fixed, just needs a rebase.

PR #476, opened 2023-07-13
Swap to regex-automata and significantly improve regex performance

CraftSpider:
Also makes lookaround and anchors work better - `^` no longer always matches, but instead literally matches starts of lines. While this removes all difference between bytes and not bytes, that shouldn't actually change any behavior - unicode mode is still enabled on bytes in the regex crate, so if anything, the distinction likely causes more confusion than it solves.

I'd like to expose some of the knobs regex-automata has to the user eventually - most pressingly, enabling/disabling unicode mode at the parser level - but I haven't found the best way to handle that. I'd like to not build the regex during parse-time, but to not do that, we'd instead have to rebuild it every time the user changed a setting, which seems a little unfortunate. On the other hand, since it happens at parser building time not parse time, I'd be willing to eat the hit if there isn't a better way.

### DIFF
--- a/src/input.rs
+++ b/src/input.rs
@@ -148,6 +148,10 @@ pub trait ExactSizeInput<'a>: Input<'a> {
 pub trait SliceInput<'a>: ExactSizeInput<'a> {
     /// The unsized slice type of this input. For [`&str`] it's `&str`, and for [`&[T]`] it will be `&[T]`.
     type Slice;
+    
+    /// Get the full slice of the input
+    #[doc(hidden)]
+    fn full_slice(&self) -> Self::Slice;
 
     /// Get a slice from a start and end offset
     // TODO: Make unsafe
@@ -250,6 +254,11 @@ impl<'a> SliceInput<'a> for &'a str {
     type Slice = &'a str;
 
     #[inline(always)]
+    fn full_slice(&self) -> Self::Slice {
+        *self
+    }
+
+    #[inline(always)]
     fn slice(&self, range: Range<Self::Offset>) -> Self::Slice {
         &self[range]
     }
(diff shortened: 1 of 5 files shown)


src/input.rs
```diff
@@ -148,6 +148,10 @@ pub trait ExactSizeInput<'a>: Input<'a> {
 pub trait SliceInput<'a>: ExactSizeInput<'a> {
     /// The unsized slice type of this input. For [`&str`] it's `&str`, and for [`&[T]`] it will be `&[T]`.
     type Slice;
+    
+    /// Get the full slice of the input
+    #[doc(hidden)]
+    fn full_slice(&self) -> Self::Slice;
```
zesterer, line 154: Hmm, I'm not sure what this adds. What's the advantage of taking the full slice and using a sub-range rather than just using the required sub-slice?

CraftSpider: This is for the aforementioned fixes to anchors - the example of what this fixes is basically the canonical example on the documentation for `span`, but to explain more: imagine the regex `\b\w+\b` - this should match only a whole word, requiring a boundary at both sides. With a parser consuming a padded `foo`, this would be expected to match `foo bar` but not `foobar` - but currently, it will match both, because slicing the input down means the regex engine sees nothing before `bar`. With this change, the engine will start matching at `bar`, but be aware of whether ` ` or `foo` directly precedes it, fixing this behavior.

zesterer: Ah, nice. That explains all, thanks :)

zesterer (approved):

zesterer:
Looks like there are some CI things to fix.

zesterer:
Thanks!

PR #481, opened 2023-07-15
Add serde feature and support

### LINKED ISSUE #479: SimpleSpan & Serde
In chumsky 0.9, I could serialise the output of my parser by `#[derive(Serialize, Deserialize)]` on my syntax tree.
Using chumsky 1.0.0-alpha.4, I cannot do this anymore, because [SimpleSpan](https://docs.rs/chumsky/1.0.0-alpha.4/chumsky/span/struct.SimpleSpan.html#) does not derive `Serialize`.
As a workaround, I tried to use `Range<usize>` as my `Span` type, because the [Span](https://docs.rs/chumsky/latest/chumsky/span/trait.Span.html#) documentation states that "Span is automatically implemented for [Range<T>](https://doc.rust-lang.org/nightly/core/ops/range/struct.Range.html)". Ok, let's try that:

~~~ rust
type Span = Range<usize>;
type Error<'a> = extra::Err<Rich<'a, char, Span>>;
~~~

But I cannot use this! Consider:

~~~ rust
fn foo<'a>() -> impl Parser<'a, &'a str, &'a str, Error<'a>> {
    todo()
}
~~~

This does not compile:

~~~
error[E0271]: type mismatch resolving `<&str as Input<'_>>::Span == Range<usize>`
~~~

Ok, so what's `<&str as Input<'_>>::Span`? [It's SimpleSpan](https://docs.rs/chumsky/1.0.0-alpha.4/chumsky/input/trait.Input.html#impl-Input%3C'a%3E-for-%26'a+str)! And there must be somewhere a constraint that dictates that the span of the input must be the same as the span type as chosen by the user.

So how can I use a `Span` such that I can serialise it?

CraftSpider:
Closes #479

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,7 +23,7 @@ std = []
 nightly = []
 
 # Allows deeper recursion by dynamically spilling stack state on to the heap.
-spill-stack = ["stacker", "std"]
+spill-stack = ["dep:stacker", "std"]
 
 # Allows parser memoization, speeding up heavily back-tracking parsers and allowing left recursion.
 memoization = []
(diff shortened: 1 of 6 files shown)


CraftSpider:
Will fix the failing tests tomorrow if I have time

zesterer:
LGTM!

zesterer (approved):

PR #482, opened 2023-07-17
SimpleSpan::eoi is much easier to understand

Lokathor:
This adds `SimpleSpan::eoi` which is I think much more specific than remembering that you can splat the length.


zesterer:
How about a method like `end_span` on the `Span` trait with signature `fn end_span(&self) -> Self;`? I think this is likely to be shorter and more general.

PR #492, opened 2023-07-27
foo-example: fix by adding 'padded'

### LINKED ISSUE #491: Example foo.rs fails to parse expr in parentheses
I was playing with the example foo.rs but it fails with the following code
```
let x = (123);
x + 123
```
-> Parse error: found "x" but expected one of "/", "+", "-", end of input, "*"

Can somebody give me a hint why this does not work? Cause otherwise parenthesed exprs just work..

Zij-IT:
Fixes #491 by adding `padded` to `sum` which ensures that all expressions allow for trailing / leading whitespace


Zij-IT:
All good 😅 I just saw your comment in the issue

Issue #486, opened 2023-07-24
Memory leak in `recursive` when using `define` function


mattnenterprise:
The `recursive` implementation leaks memory if it references itself and the parser definition is defined using the `define` function. See the following example that creates millions of parsers, but the memory is never released.

```rust
use chumsky::prelude::*;

#[derive(Debug, PartialEq)]
enum Chain {
    End,
    Link(char, Box<Chain>),
}

fn parser() -> impl Parser<char, Chain, Error = Simple<char>> {
    let mut chain = Recursive::<_, _, Simple<char>>::declare();

    chain.define(just('+')
        .then(chain.clone())
        .map(|(c, chain)| Chain::Link(c, Box::new(chain)))
        .or_not()
        .map(|chain| chain.unwrap_or(Chain::End)));

    chain
}

fn main() {
    for _n in 1..100_000_000 {
        parser();
    }
}
```

I originally found this issue in the jaq filter parser [here](https://github.com/01mf02/jaq/blob/main/jaq-parse/src/filter.rs#L210-L255)

zesterer:
Hi,

It seems like `jaq` is using `0.9`. The `1.0` release is a from-scratch rewrite of chumsky. Do you experience the same issue there?

mattnenterprise:
`jaq` generates lots of compile errors when I try to upgrade from `0.9` to `1.0`, and I'm not knowledgeable enough about the code base to upgrade them all.

I was able to reproduce the memory issue with the most current commit on main 6bb31a5f070e1289a11ea4a0193c7ccb236c81fa. I did so with the following code.

```rust
use chumsky::prelude::*;

#[derive(Debug, PartialEq)]
enum Chain {
    End,
    Link(char, Box<Chain>),
}

fn parser<'a>() -> impl Parser<'a, &'a str, Chain, extra::Err<Simple<'a, char>>> {
    let mut chain = Recursive::declare();

    chain.define(just::<_, _, extra::Err<Simple<char>>>('+')
        .then(chain.clone())
        .map(|(c, chain)| Chain::Link(c, Box::new(chain)))
        .or_not()
        .map(|chain| chain.unwrap_or(Chain::End)));

    chain
}

fn main() {
    for _n in 1..100_000_000 {
        parser();
    }
}
```
I actually based this example on an example in the recursive documentation [here](https://github.com/zesterer/chumsky/blob/6bb31a5f070e1289a11ea4a0193c7ccb236c81fa/src/recursive.rs#L102)

[...]

PR #494, opened 2023-07-28
recursive-leak: fix leak by using middle-man

Zij-IT:
This introduces a middle man to prevent both memory leaks (#486) in recursive parsers, and prevents the user from foot-gunning themselves by calling the wrong `clone` (referring to craftspider's `weak_clone`) suggestion.


src/recursive.rs
```diff
@@ -128,24 +147,35 @@ impl<'a, 'b, I: Input<'a>, O, E: ParserExtra<'a, I>> Recursive<Indirect<'a, 'b,
...
 
+impl<'a, 'b, I: Input<'a>, O, E: ParserExtra<'a, I>> Recursive<Indirect<'a, 'b, I, O, E>, Defined> {
     /// Defines the parser after declaring it, allowing it to be used for parsing.
     // INFO: Clone bound not actually needed, but good to be safe for future compat
     #[track_caller]
-    pub fn define<P: Parser<'a, I, O, E> + Clone + MaybeSync + 'a + 'b>(&mut self, parser: P) {
+    pub fn define<P: Parser<'a, I, O, E> + Clone + MaybeSync + 'a + 'b>(
+        parser: P,
+        declaration: Recursive<Indirect<'a, 'b, I, O, E>, Declared>,
+    ) -> Self {
```
zesterer, line 162: Should this function be on `Recursive<..., Declared>` and return a `Recursive<..., Defined>`?

Zij-IT: Yeah, you could do that too. For type inference it doesn’t matter, but it could make more sense logically.

zesterer: I'd say so, yes.

CraftSpider:
While this method seems nice, it also makes more complicated recursive patterns harder - you can't have a single context struct on which you both use and define the recursive parser, since the two things are separate types. While you may be able to re-architect code in most cases to avoid this, I'm curious if that's an intended limitation?

Zij-IT:
> While this method seems nice, it also makes more complicated recursive patterns harder - you can't have a single context struct on which you both use and define the recursive parser, since the two things are separate types. While you may be able to re-architect code in most cases to avoid this, I'm curious if that's an intended limitation?

No, it's not intended. We could wrap them both in an enum (somewhat like `Cow`), and have an arm for both `Defined` and `Declared`, which would mean that the user would only have to deal with `Declared` and `Defined`. The one thing I don't like about the enum is that it opens the door for the user to attempt to use `define` on a parser that is already been defined, which isn't possible with the way it's currently structured.

src/recursive.rs
Zij-IT: Completed with 4572bb5

Zij-IT:
Given that those checks fail for reasons outside of this PR, how would you like me to proceed?

zesterer:
You should be able to rebase to fix the failing CI now :)

Zij-IT:
Finally got around to it. My bad for the mess that is the commit history in this pull request :sweat_smile: One would think that after using it for four+ years that I would be able to keep it somewhat clean

zesterer:
I totally forgot about this PR, I'll take a final look at it tomorrow.

zesterer:
Was the issue mentioned by CraftSpider above resolved?

faassen:
Is there any way I could help on this issue?

zesterer:
I think it probably just needs a rebase and then it's ready to go. That said, I am a bit concerned about the issue @CraftSpider mentioned above: it will probably break the way some people are using `Recursive`. That said, arguably such uses are just wrong and *should* be discouraged?

CraftSpider:
I _think_ that you can probably change it to always store one or the other in the state, but I haven't really gotten around to trying it out with this change. I'm not too worried personally.

zesterer:
Right, in that case it's probably fine. I would appreciate the commit history being cleaned up a bit here before merging too (otherwise I think I'll squash).

faassen:
I need to look at my use of recursive tomorrow to see whether whether I can understand the limitation and whether it hurts my usage, I don't get it yet without staring at the code.

zesterer:
I believe it's specifically for the case of `Recursive::declare`/`Recursive::define`. Doing `my_recursive.define(...)` previously had the signature `fn(self: &mut Recursive, impl Parser)` (i.e: it was an in-place mutation) but now has the signature `fn(self: Recursive<A>, impl Parser) -> Recursive<B>` (i.e: it consumes `self` and transforms it to a new type, thereby turning it into a strong reference, as is the case internally with the `recursive` function).

Zij-IT:
So, my apologies for dropping off of the face of this earth for so long. If I get this rebased in the next coming days, would it be good to be merged, or is there anything else that it is needing?

zesterer:
I think this change is probably the way to go, on reflection. It's unfortunate that it breaks certain uses of `Recursive::declare/define` but those modes of use are *inherently* leaky in a way that seems like it would require generalised garbage collection to disentangle. So, yes, I'd be happy to see this merged.

Zij-IT:
So... I'll be honest. I tried to rebase this thing for like an hour :') Merging took all of two seconds, so obviously I suck at rebasing.

wackbyte:
I think the fix for the failure is just replacing ``[`Recursive`](Recursive)`` with ``[`Recursive`]``

zesterer:
Maybe it would be better to just take the diff from `main`, then go back to `main` and apply that diff on top, creating a new commit entirely? Then you can force-push that to this branch.

Flygrounder:
@zesterer Could we have a new alpha release once this gets merged?

Flygrounder:
> Maybe it would be better to just take the diff from `main`, then go back to `main` and apply that diff on top, creating a new commit entirely? Then you can force-push that to this branch.
> 
> Don't worry about unrelated CI stuff, I'll address that elsewhere :)

Wanted to ask, isn't it the same as using "Squash and merge" option when merging on GitHub? It seems like @Zij-IT got busy and it's a bit sad that this PR is essentially ready, but not merged. We want to use this library in production in our company, but this memory leak is a problem in our use case (long-running process that can parse thousands of inputs).

Zij-IT:
My apologies to everyone that has been waiting for me to make the change. I just did what was recommended by @zesterer, so this should be good to roll.

The test that fails fails because what was a runtime error became a compiler error in the tests case. There isn't AFAIK a way to test that something fails to compile, so I can't write an equivalent test. As such, I just deleted the test :/

wackbyte:
You can test something that fails to compile by writing a doctest:
```rs
/// ```compile_fail
/// // ...
/// ```
fn test_name() {}
```
A specific error code can be provided, separated by a comma: `compile_fail,E9999`.

There might be some more setup required, but this is a method I've used in the past.

This has the drawback of not being able to specify an expected error. A better solution would likely use `trybuild`.

src/recursive.rs
```diff
@@ -128,24 +147,36 @@ impl<'a, 'b, I: Input<'a>, O, E: ParserExtra<'a, I>> Recursive<Indirect<'a, 'b,
...
     }
 
     /// Defines the parser after declaring it, allowing it to be used for parsing.
     // INFO: Clone bound not actually needed, but good to be safe for future compat
     #[track_caller]
-    pub fn define<P: Parser<'a, I, O, E> + Clone + MaybeSync + 'a + 'b>(&mut self, parser: P) {
+    pub fn define<P: Parser<'a, I, O, E> + Clone + MaybeSync + 'a + 'b>(
+        parser: P,
+        declaration: Self,
+    ) -> Recursive<Indirect<'a, 'b, I, O, E>, Defined> {
```
zesterer, line 160: What's the reason for not having the `declaration` argument be `self`?

Zij-IT: Honestly nothing in particular. I'll swap it over to having it be a method!

Zij-IT: @zesterer Which one do you prefer? I'll do it whichever way you'd prefer. I personally like the symmetry between `Recursive::declare` and `Recursive::define` and the second way makes it more obvious that `Recursive::define` consumes the parser.

zesterer: Hmm, I think I prefer the first, personally. The extra parameter way off at the end of the definition feels very isolated and random if you're not already sure what it is. Also, putting it at the front makes clear what the target of the definition is.

Zij-IT:
I'm pretty sure GitHub decided the 7 commits before mine are different because the hashes are different... I try and do the right thing with a rebase I manage to create clutter :D

Anyhow, this should be the final commit for the PR to be merged, right?

zesterer:
I'll try to find time to take another pass of this today. Hopefully, yep!

Zij-IT:
So, I don't believe this solution will work in the way that I was hoping. It was intended to help situations like seen [here](https://github.com/lakehq/sail/blob/main/crates/sail-sql-parser/src/parser.rs).

However, due to the issue described jkylling in #664. I found a test case in the wild by Sail's SQL Parser. If you apply this [patch](https://github.com/user-attachments/files/23142023/recursive.patch) to the lastest version of Chumsky, and run the tests for [sail-sql-analyzer](https://github.com/lakehq/sail/tree/main/crates/sail-sql-analyzer), it will fail due to a parser "not being defined before it's used". The definition of the failing parser is [here](https://github.com/lakehq/sail/blob/main/crates/sail-sql-parser/src/parser.rs#L92-L119).

I haven't tested if the solution in #664 can be extended to support $N$ recursive parsers as seen in the sail parser without. If so, it would be great as it prevents the foot-gun of holding onto a parser for too short a time, which this solution clearly does not.

Here is the patch for the Sail parser: [patch](https://github.com/user-attachments/files/23142101/parser.patch)

PR #497, opened 2023-08-02
docs: link examples & tutorial of right version on docs.rs

not-my-profile:
Previously the start page of the documentation on docs.rs for chumsky linked the examples and tutorial from the main branch, no matter which version you selected on docs.rs.

So e.g. https://docs.rs/chumsky/0.9.2/chumsky/ links https://github.com/zesterer/chumsky/blob/main/examples/brainfuck.rs which is confusing since the main branch currently contains the source code for the upcoming 1.0.0, which has a different API.

This commit remedies this for future releases by separating the link labels and URLs using CommonMark link reference definitions[1] and overriding them in src/lib.rs, building the correct URLs via CARGO_PKG_VERSION[2].

This trick will require git tags to be created for all future versions (so e.g. if 0.9.3 will be released the commit will have to be tagged with "0.9.3").

[1]: https://spec.commonmark.org/0.30/#link-reference-definition
[2]: https://doc.rust-lang.org/cargo/reference/environment-variables.html

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,8 +42,7 @@ A parser library for humans with powerful error recovery.
 
 ## Example [Brainfuck](https://en.wikipedia.org/wiki/Brainfuck) Parser
 
-See [`examples/brainfuck.rs`](https://github.com/zesterer/chumsky/blob/master/examples/brainfuck.rs) for the full
-interpreter (`cargo run --example brainfuck -- examples/sample.bf`).
+See [`examples/brainfuck.rs`] for the full interpreter (`cargo run --example brainfuck -- examples/sample.bf`).
 
 ```rust
 use chumsky::prelude::*;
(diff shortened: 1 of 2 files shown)


not-my-profile:
I just realized that `tutorial.md` has several links to docs.rs so it would probably best to move it into a module within the crate so that the links point to the right version. An example of a project that does this is Polars (see [polars::docs](https://docs.rs/polars/0.31.1/polars/docs/)). However I'd rather leave this to a followup PR (if you see the merit).

zesterer:
Thanks! Do you have any external information I could read about this technique?

Feel free to ignore `tutorial.md` for now, it's due to be replaced as part of the new guide anyway :)

not-my-profile:
The technique relies on former link definitions taking precedence over latter link definitions which is specified by CommonMark, see [example 204](https://spec.commonmark.org/0.30/#example-204). Rustdoc uses the `pulldown_cmark` Markdown parser which is CommonMark compliant so this works.

>  [tutorial.md is] due to be replaced as part of the new guide anyway :)

Ah right ... I see that the new guide does exactly what I suggested (being part of the crate) ... so that's great :)

zesterer (approved):
Thanks!

zesterer:
Thanks!

PR #502, opened 2023-08-17
Implement choice parser for vec

### LINKED ISSUE #484: Feature Request: Implement ParserSealed for Choice<Vec<A>>
I want to build a choice parser at runtime, using some user-supplied tokens.
However, it doesn't seem like that's possible with the latest chumsky version
https://github.com/zesterer/chumsky/blob/10dad10f6a9254456bc56b91d2f0763427d7ddec/src/primitive.rs#L841

I could instead use a number of chained .or() parsers, but that might not be an ideal solution.

Background info why I even want this in the first place:
I'm writing an extensible math parser. Extensible, because I cannot possibly cover all areas of mathematics. At the very least, users should be able to define new symbols, like saying `lim` is a prefix operator and not a funnily named variable. Which would translate to `choice(vec![lim parser, variable parser, other parsers])`

stefnotch:
Fix #484 

I hope I got the implementation down right. It feels a bit odd to just duplicate the code, but I'll take it.

I can also add a test for it, if there's a place where you think it'd be useful.

### DIFF
--- a/src/primitive.rs
+++ b/src/primitive.rs
@@ -872,38 +872,7 @@ macro_rules! impl_choice_for_tuple {
 
 impl_choice_for_tuple!(A_ B_ C_ D_ E_ F_ G_ H_ I_ J_ K_ L_ M_ N_ O_ P_ Q_ R_ S_ T_ U_ V_ W_ X_ Y_ Z_);
 
-impl<'a, A, I, O, E, const N: usize> ParserSealed<'a, I, O, E> for Choice<[A; N]>
-where
-    A: Parser<'a, I, O, E>,
-    I: Input<'a>,
-    E: ParserExtra<'a, I>,
-{
-    #[inline]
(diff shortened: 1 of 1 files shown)


zesterer:
Awesome! If you wanted to avoid the duplication, a neat way to do this would be to implement `Parser` for `Choice<&'b [A]>` and then have the implements for `Choice<Vec<A>>` and `Choice<[A; N]>` both call out to it (the implementation can literally just be `choice(&self[..]).go(inp)` in both cases, I think). What do you think?

stefnotch:
@zesterer Okay, I gave it a shot. Is this what you meant?

zesterer:
Yep, exactly that! Thanks very much :)

stefnotch:
Oh no, Clippy is rightfully upset with my code. Fixed it.

zesterer:
Awesome :)

Issue #354, opened 2023-03-07
Hide implementation details within sealed trait(s)


zesterer:
Some things me might want to hide:

- `Mode`
- `Check`/`Emit`/etc.
- `Parser::go`/`Parser::go_check`/`Parser::go_emit`/etc.
- `InputRef`
- Most of `Input` (`Input::next_maybe`, for example)

Hiding these details does not necessarily make the crate non-extensible! A viable route to recovering extensibility is to have a more simplistic set of traits for common interfaces (like `Parser`, `Input`, etc.) that provide a more limited API for which we can guarantee long-term stability, and then provide blanket implementations of the more complex sealed traits for implementers of the simpler traits.

The goal here is *not* to hurt downstream developers that may want to build upon chumsky, but to protect them from changes to volatile implementation details that are likely to happen as we find more ways to aggressively optimise parsers. This has a twofold benefit:

1) Developers can extend the trait without fear that the ground is suddenly going to shift under them, providing room for an ecosystem to build up around a stable core interface

2) We, as developers of said core, get the change to aggressively optimise the core and alter the way in which parsers work to provide users with better performance without worrying that we're hurting downstream extensions

stefnotch:
One thing that I would find interesting would be a combinator that takes multiple parsers, and "merges" them.
It would apply all parsers to an input, and return whichever one leads to a longer result.

This would be convenient for writing grammars in really simple ways and not having to manually merge them correctly. The boring example would be
```
Statement → 
  "if" Condition "{" Statement "}" |
  "if" Condition "{" Statement "}" else "{" Statement "}"
```

While the way more interesting example is parsing decimal and hexadecimal numbers. 
```
Number → 
  ("0" | "1" | "2" | ...)+ |
  "0x" ("0" | "1" | ... | "a" | "b" ...)+
```

The or combinator is similar to that, but it has a different behavior.

zesterer:
That's an interesting idea. We'd have to be careful to avoid exponential blowup (if you nest such parsers, even the happy path becomes exponential). Sort of like `choice`, except the parser that left the input the furthest progressed wins.

This is absolutely possible with the extension API I implemented today, and actually quite easy to implement.

PizzaCrust:
It would be really nice to have `Mode` publicly available instead of splitting it into two functions.
I don't really want to copy the same code twice just without the output otherwise leaving potential performance on the table

zesterer:
> It would be really nice to have `Mode` publicly available

[...]

PR #511, opened 2023-08-26
Implement Ord for Offset struct

stefnotch:
Was nice and straightforward, since the [underlying Offset](https://github.com/zesterer/chumsky/blob/f10e56b7eac878cbad98f71fd5485a21d44db226/src/input.rs#L36) is already required to be Ord.

See also https://github.com/zesterer/chumsky/issues/354#issuecomment-1694332577


zesterer:
Thanks so much! This is ideal :)

Issue #399, opened 2023-04-13
Feature request: pipelined parsers


Banyc:
Say we have the two parsers:

- `fn lexer() -> impl Parser<char, Vec<Token>, Error = Simple<char>>;`
- `fn token_parser() -> impl Parser<Token, Expr, Error = Simple<Token>>;`

# Feature request

A way to merge the two parsers as one.

```rust
let parser = token_parser().inputed_from(lexer());
parser.parse(src)
```

# Potential benefits

- Prevent the creation of `Vec<Token>` which takes up a bit of heap memory, so that it makes inter-procedures zero-copy.
- `parser.parse(src)` seems cooler than `let tokens = lexer.parse(src); token_parser.parse(tokens)` in user code. By "users" I mean those who use the built parser(s).

# My questions

1. Is it possible to do this with current implementations?
1. Is the idea makes sense to most use cases so that it worths being implemented?

Banyc:
Btw, thank you for making this project and the Veloren game, which make my at-work and off-work hours easier/happier. Also the tutorials of the two projects are great!

Banyc:
I haven't got the full picture on the Rust ecosystem so I probably wrong, but I have some idea on the first one:

1. I want to mathematically reduce the `next` calling chain into something like this:
   ```text
         /    /
        /    /
   ---->|-+->|-+
        | |  | |
   <----|-+  | |
        |    | |
   <----|----|-+
        \    \
         \    \
   ```
   - Each layer is a `next` call
1. `tower::Service` can also be reduced into this model
1. `tower::Service` takes a `Request` as input and `Response` as output. Error type is usually `Box<dyn Error>`
   - the [official guide](https://github.com/tower-rs/tower/blob/master/guides/building-a-middleware-from-scratch.md) recommends to use `Box<dyn Error>` as the propagating error type
   - the guide reasons about why to do so
1. Therefore, it might be a good idea to `Box<dyn std::error::Error>` the error

zesterer:
In your diagram, what do those distinct outflows represent? Would errors from the inner parser (i.e: the lexer, in most cases) get propagated up to become parser errors?

zesterer:
[...]

PR #513, opened 2023-08-28
enable using a tokenized input as a Stream input

cosmicexplorer:
### Problem

I'm parsing a string into tokens with chumsky, and I would like to also use chumsky to parse those tokens into something else. While `select! { ... }` is intended to enable this, it assumes that the stream of tokens is produced externally to chumsky, as in the logos example: https://github.com/zesterer/chumsky/blob/56762fe5c965880066a068038a18b5ac07afd75c/examples/logos.rs#L130-L145

### Solution
- Expose `.parse_iter()` outside of `#[cfg(test)]` and use it to construct a `Stream` instance.
- Expose `.stream(input)` as a public method of `IterParser` to generate a stream of transformed input.

### DIFF
--- a/src/input.rs
+++ b/src/input.rs
@@ -1033,7 +1033,6 @@ where
         }
     }
 
-    #[cfg(test)]
     pub(crate) fn as_ref_at<'parse>(
         &'parse mut self,
         offset: I::Offset,
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2306,7 +2306,6 @@ where
 }
 
 /// An iterator that wraps an iterable parser. See [`IterParser::parse_iter`].
-#[cfg(test)]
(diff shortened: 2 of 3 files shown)


zesterer:
One thing I worry about is that the API seems to imply that the parser gets turned into a `Stream`, when in reality it's used used to parse elements, collected into a vector, and then those elements are used as a `Stream`. #399 discusses the former use-case and what problems we've run up against when trying to do this.

Did you have an example of the sort of patterns that this enables?

cosmicexplorer:
Ah, I see #399 (mentioned directly above `parse_iter()`) covers exactly this issue, not a different one. I'll see if I can page into that.

cosmicexplorer:
Converted this into a draft as this is really just the easier part of #399.

PR #514, opened 2023-08-28
Pass operator to Pratt parsing functions

### LINKED ISSUE #507: [Feature request] Pratt parsing with spans
When using the new pratt parsing feature, how would I create a syntax tree which includes spans?

e.g. How does one get spans that include the operators? And how does one get the spans of the operators themselves?
https://github.com/zesterer/chumsky/blob/01b96cd643e4b47b36ee76baa57294d2153965f7/src/pratt.rs#L709-L719

stefnotch:
Deprecated in favor of https://github.com/zesterer/chumsky/pull/515

Fix #507 
I'm not sure if this implementation is ideal, so I'll appreciate any and all feedback. 
Things that could definitely still be done
- An additional test, where this feature is actually being used to get a span.
- Benchmarking, if useful.

Pinging @Zij-IT : I'd love to get some feedback on this.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,7 +15,7 @@ exclude = [
 build = "build.rs"
 
 [features]
-default = ["std", "spill-stack"]
+default = ["std", "spill-stack", "pratt"]
 
 # Integrate with the standard library.
 std = [
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,7 +12,6 @@ exclude = [
     "/misc/*",
 	"/benches/samples/*",
 ]
-build = "build.rs"
 
 [features]
 default = ["std", "spill-stack"]
@@ -92,9 +91,6 @@ sn = "0.1"
 logos = "0.13"
 lasso = "0.7"
 slotmap = "1.0"
-
-[build-dependencies]
-vergen = { version = "=8.1.1", features = ["git", "gitoxide"] }
 
 [target.'cfg(unix)'.dev-dependencies]
 pprof = { version = "0.11", features = ["flamegraph", "criterion"] }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,7 +15,7 @@ exclude = [
 build = "build.rs"
 
 [features]
-default = ["std", "spill-stack", "pratt"]
+default = ["std", "spill-stack"]
 
 # Integrate with the standard library.
 std = [
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2174,12 +2174,12 @@ pub trait Parser<'a, I: Input<'a>, O, E: ParserExtra<'a, I> = extra::Default>:
     /// let operator = choice((
     ///     // Our `-` and `+` bind the weakest, meaning that even if they occur
     ///     // first in an expression, they will be the last executed
-    ///     left_infix(just('+'), 1, |l, r| Expr::Add(Box::new(l), Box::new(r))),
-    ///     left_infix(just('-'), 1, |l, r| Expr::Sub(Box::new(l), Box::new(r))),
+    ///     left_infix(just('+'), 1, |_op, [l, r]| Expr::Add(Box::new(l), Box::new(r))),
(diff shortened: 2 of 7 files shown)


zesterer:
@Zij-IT Any chance you'd have a free moment to take a look at this?

Zij-IT:
> @Zij-IT Any chance you'd have a free moment to take a look at this?

I’ll give it a look this weekend (:

Cargo.toml
```diff
@@ -15,7 +15,7 @@ exclude = [
 build = "build.rs"
 
 [features]
-default = ["std", "spill-stack"]
```
Zij-IT: I believe this gets fixed in a following commit, but tagging just in case :)

src/lib.rs
```diff
@@ -2208,14 +2208,13 @@ pub trait Parser<'a, I: Input<'a>, O, E: ParserExtra<'a, I> = extra::Default>:
...
+    fn pratt<InfixParser, InfixOperator>(
         self,
-        ops: InfixOps,
-    ) -> Pratt<I, O, E, Self, pratt::Infix<InfixOps, InfixOpsOut>>
+        ops: InfixParser,
+    ) -> Pratt<I, O, E, Self, pratt::Infix<InfixParser, InfixOperator>>
     where
         I: Input<'a>,
         E: ParserExtra<'a, I>,
-        InfixOps: Parser<'a, I, InfixOpsOut, E>,
```
Zij-IT, line 2218: Any particular reason this trait bound has to disappear? It was originally added with the idea that it makes no sense to construct a `Parser` which cannot parse.

src/pratt.rs
```diff
@@ -213,31 +235,30 @@ impl<'a, I, O, E, Atom, InfixOps, InfixOpsOut, PrefixOps, PrefixOpsOut>
...
+            PrefixParser,
+            PrefixOperator,
+            PostfixParser,
+            PostfixOperator,
         >,
     >
     where
         I: Input<'a>,
         E: ParserExtra<'a, I>,
-        PrefixOps: Parser<'a, I, PrefixOpsOut, E>,
```
Zij-IT, line 240: Any particular reason this trait bound has to disappear? It was originally added with the idea that it makes no sense to construct a Parser which cannot parse.

src/pratt.rs
```diff
@@ -252,11 +273,11 @@ impl<'a, I, O, E, Atom, InfixOps, InfixOpsOut, PostfixOps, PostfixOpsOut>
...
 }
 
-type InfixBuilder<E> = fn(lhs: E, rhs: E) -> E;
+type InfixBuilder<Op, O> = fn(op: Op, children: [O; 2]) -> O;
 
-type PrefixBuilder<E> = fn(rhs: E) -> E;
+type PrefixBuilder<Op, O> = fn(op: Op, child: O) -> O;
 
-type PostfixBuilder<E> = fn(rhs: E) -> E;
+type PostfixBuilder<Op, O> = fn(op: Op, child: O) -> O;
```
Zij-IT, line 280: This is one of the points that I was unsure about. I think I would prefer and find it more natural for `PostfixBuilder` to be `fn(O, Op) -> O` because the op follows the expression when parsing.

Zij-IT (commented):
Overall I love the work, and I think you did a great job. @zesterer has a similar PR that also moves us closer to the API using `Fn` as opposed to the current API which uses `fn`, and also allows us to use multiple combinations of functions, such as the one you would prefer, and also one that has a total span.

I'm curious what you would think of his variation, and if you feel there are any items missing in that implementation? If you haven't seen it, you can find it here #515

src/lib.rs
stefnotch: It doesn't have to disappear, but it does have to change a bit. At first I didn't figure out how to change it, so I haphazardly deleted it.

stefnotch:
Oh sweet, zesterer might have built an even more general version. I'll check it out and am very happy with closing this in favour of a better solution!

As long as I end up being able to use an excellent parser combinator library, I'm very satisfied.

stefnotch:
Deprecated in favor of https://github.com/zesterer/chumsky/pull/515

PR #517, opened 2023-09-06
WIP: Add map_with_ctx

jess-sol:
While building a parser for yaml I found a method like this necessary. The context for my parser looks something like this:
```rust
pub struct ParsingContext {
    pub(crate) block_key: bool,
    pub(crate) in_block: bool,
    pub(crate) in_flow: bool,
    indent: Option<usize>,
}
```

Which is [necessary context](https://yaml.org/spec/1.2.2/#rule-s-separate) for parsing. Only one of these fields needs to be updated at a time, and normally this would be easy enough with `map_ctx`, but when trying to update indentation using a parser & `ignore_with_ctx`, it becomes a problem.

This could be solved a few ways directly:
1. A `map_with_ctx` function allowing the merging of existing and new context, which will then likely be used by a `*_with_ctx` call.
2. Add new `*_with_ctx` methods which support modifying the context being used
3. Reconfigure `*_with_ctx` to append to a tuple, instead of replacing existing context entirely. This would mean context would nest.

An example of it's usage:
```rust
    text::inline_whitespace()
        .configure(|cfg, ctx: &ParsingContext| cfg.at_least(ctx.indent.map(|x| x + 1).unwrap_or(0)))
        .count()
        .rewind()
        .map_with_ctx(|ind, ctx: &ParsingContext| { let ctx = ctx.clone(); ctx.indent = Some(ind); ctx })
        .ignore_with_ctx(line_parser)
```


src/private.rs
```diff
@@ -87,6 +95,14 @@ impl Mode for Emit {
         f(x)
     }
     #[inline(always)]
+    fn map_with_ctx<T, U, C, F: FnOnce(T, C) -> U>(
+        x: Self::Output<T>,
+        ctx: C,
+        f: F,
+    ) -> Self::Output<U> {
+        f(x, ctx)
+    }
```
zesterer, line 104: I think you should be able to do this with just `M::map(x, move |x| f(x, c))`, so no need to add another method to `Mode`.

src/lib.rs
```diff
@@ -525,6 +525,31 @@ pub trait Parser<'a, I: Input<'a>, O, E: ParserExtra<'a, I> = extra::Default>:
...
+    /// current context.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// # use chumsky::{prelude::*, error::Simple};
+    ///
+    /// // let x = empty::<char, _>().to(0usize).map_with_ctx(|ind: usize, ctx: &()| -> usize { ind });
+    /// // let indent = empty::<&str, extra::Default>().to(0usize).map_with_ctx(|ind: usize, ctx: ()| ind);
+    /// let indent = text::whitespace::<char, &str, extra::Default>().count(); //.map_with_ctx(|ind: usize, ctx: usize| ctx + ind);
```
zesterer, line 538: I'm not sure I understand this example.

zesterer (commented):
Thanks! I think this PR seems reasonable, although I'd be interested to hear more about the use-case for this. All use-cases I can think of can be covered by existing combinators.

src/lib.rs
jess-sol: I didn't actually finish the example, I wanted to make sure you were interested in including a method like this before I spent time on the docs. I can clean it up before removing WIP though, definitely

jess-sol:
Yeah, I wonder the same honestly. I'm no expert in Chumsky, but I did hit a wall with this one. [Here's an example][1] of where I think it'd be necessary. When setting the current indentation level, the current context needs to be merged with the output of a parser.

I didn't see another way of doing that, short of `custom` which would obviously work here. I assume Chumsky would want to be able to parse yaml eventually with no use of custom; so I planned on working towards that eventually.

[1]: https://gist.github.com/jess-sol/b354f623c3b79c61e0465eef0e595847#file-lib-rs-L125

jess-sol:
I took another stab at the docs for `map_with_ctx`, though I'm not entirely happy with the example yet. It might also be good to add a `try_map_with_ctx` for completeness.

I'm struggling to come up with a simpler [example than this](https://gist.github.com/jess-sol/f4eb75eea4eebac11cadce7c38413834) that isn't contrived. I'm trying to demonstrate the issue I'm facing with parsing YAML using Chumsky: parsing bare strings (and to a lesser extent all scalars) is very context dependent. It's not feasible right now without using custom.

Parsing YAML is very different than any of the other examples I've seen Chumsky used for. But I think it's a perfect use-case for the context system; which I think could easily be made to be *just* powerful enough to do the job.

zesterer:
Just to let you know, this is absolutely on my list of things to fully review, I'll hopefully get to it within the next few days.

zesterer:
Ah, I can see the value here. So it looks like this is a way to add information to the context for specific patterns?

zesterer (approved):

zesterer:
Looks like this might need a rebase.

zesterer:
Looks like the doc test has a problem.

jess-sol:
Ah yeah, because tests are setup with all features enabled, `sync` was breaking because the `dyn Parser` wasn't `Send + Sync`.

PR #520, opened 2023-09-13
Bind span earlier in `select{_ref}!` so that it may be used in the `if` guard

wackbyte:
(Suggested by @CraftSpider)


zesterer:
Nice, this is a good change. CI failures are unrelated, so will merge. Thanks!

PR #528, opened 2023-09-20
Add any_ref()

tesujimath:
Works just like any() but for BorrowInput rather than ValueInput.

### DIFF
--- a/src/primitive.rs
+++ b/src/primitive.rs
@@ -627,7 +627,9 @@ where
 
 /// A parser that accepts any input (but not the end of input).
 ///
-/// The output type of this parser is `I`, the input that was found.
+/// The output type of this parser is `&'a I::Token`, the input that was found.
+///
+/// This function is the borrowing equivalent of [any]. Where possible, it's recommended to use [any] instead.
 ///
 /// # Examples
 ///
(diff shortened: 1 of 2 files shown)


src/primitive.rs
```diff
@@ -591,6 +591,62 @@ pub const fn any<'a, I: Input<'a>, E: ParserExtra<'a, I>>() -> Any<I, E> {
...
+            }
+        }
+    }
+
+    go_extra!(&'a I::Token);
+}
+
+/// A parser that accepts any input (but not the end of input).
+///
+/// The output type of this parser is `I`, the input that was found.
```
zesterer, line 630: I think this should read `&'a I::Token` instead of `I`.

It might also be worth adding something like this:

> This function is the borrowing equivalent of [`any`]. Where possible, it's recommended to use [`any`] instead.

zesterer (changes requested):
Thanks for the PR! I definitely agree that this is a good idea (it's a nice parallel to `select!`/`select_ref!`). Just one small comment regarding docs.

src/primitive.rs
tesujimath: Good comment thanks, and fixed!

zesterer:
Thanks!

PR #529, opened 2023-09-20
Relax ValueInput constraint for SkipUntil and SkipThenRetryUntil

tesujimath:
Removing ValueInput constraint allows to use these with e.g. BorrowInput.

There was nothing here that required ValueInput over underlying Input.


zesterer (approved):
Thanks, definitely a welcome improvement!

PR #533, opened 2023-09-26
Only enable vergen on docsrs

### LINKED ISSUE #503: Heavyweight vergen
So I updated my chumsky, and noticed that my lockfile suddenly gained around 1000 new lines. 
Apparently vergen causes the extra dependencies. https://github.com/zesterer/chumsky/commit/136f882b46cdcbd08354a98210da86d322383b84

image

I'm not worrying too much about it, since the (from scratch) compile times for me are still pretty fast, but I figured I'd let you know.

stefnotch:
Hopefully fixes #503 

I personally would prefer if rustdoc fully supported including example code in the documentation, but this is better than nothing for now.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -59,6 +59,9 @@ regex = ["dep:regex-automata"]
 # Enable serde serialization support
 serde = ["dep:serde"]
 
+# Enable dependencies only needed for generation of documentation on docs.rs
+docsrs = ["dep:vergen"]
+
 # An alias of all features that work with the stable compiler.
 # Do not use this feature, its removal is not considered a breaking change and its behaviour may change.
 # If you're working on chumsky and you're adding a feature that does not require nightly support, please add it to this list.
@@ -94,7 +97,7 @@ lasso = "0.7"
 slotmap = "1.0"
 
 [build-dependencies]
-vergen = { version = "=8.1.1", features = ["git", "gitoxide"] }
+vergen = { version = "=8.1.1", optional = true, features = ["git", "gitoxide"] }
 
 [target.'cfg(unix)'.dev-dependencies]
 pprof = { version = "0.11", features = ["flamegraph", "criterion"] }
--- a/build.rs
+++ b/build.rs
@@ -1,8 +1,20 @@
 use std::error::Error;
+#[cfg(feature = "docsrs")]
 use vergen::EmitBuilder;
 
 fn main() -> Result<(), Box<dyn Error>> {
+    emit_git_metadata()?;
+    Ok(())
+}
+
+#[cfg(feature = "docsrs")]
+fn emit_git_metadata() -> Result<(), Box<dyn Error>> {
     // Emit the instructions
     EmitBuilder::builder().all_git().emit()?;
+    Ok(())
+}
+
+#[cfg(not(feature = "docsrs"))]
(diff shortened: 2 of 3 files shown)


Cargo.toml
```diff
@@ -59,6 +59,9 @@ regex = ["dep:regex-automata"]
 # Enable serde serialization support
 serde = ["dep:serde"]
 
+# Enable dependencies only needed for generation of documentation on docs.rs
+docsrs = ["dep:vergen"]
```
zesterer, line 63: Have you checked that this actually gets enabled with generating docsrs docs?

stefnotch: I haven't tested it with the live docsrs, since I'm not sure how one would go about doing so.

My understanding is that the `docsrs` feature gets enabled by
https://github.com/zesterer/chumsky/blob/af4e2b12cf1447e0e321af80040975254a4aa6a4/Cargo.toml#L67-L69

zesterer: I don't think this would enable such a feature. The cfg flag just allows you to do `#[cfg(docsrs)]` (which would probably be sufficient for your purposes, fwiw, instead of adding a cargo feature).

stefnotch: Okay, I tried a few things in that direction, but haven't quite managed to get it to work as expected.

So I first tested the code with `RUSTFLAGS="--cfg docsrs" cargo doc --open`. That only seems to pass the info to the Cargo.toml part. It doesn't pass the info to the `lib.rs` file, since  `#[cfg(docsrs)]` in `lib.rs` had no effect.

Then I tested it with `cargo rustdoc --open -- --cfg docsrs`. This gets the lib.rs file to cooperate, but it fails to do any of the vergen stuff.

I could probably find a variation where I get it to work, but it might almost be easier to make the `docsrs` stuff a feature.

zesterer: Have you tried the command at the top of `guide.rs`?

```
RUSTDOCFLAGS="--cfg docsrs" cargo +nightly doc --all-features
```

stefnotch: I missed that command, thank you.

So I tested it just now:
Conditionally compiling vergen does not seem to be possible with
```
[target.'cfg(docsrs)'.build-dependencies]
vergen = { version = "=8.1.1", features = ["git", "gitoxide"] }
```

Conditionally compiling it with `--all-features` and then `docsrs = ["dep:vergen"]` does seem to work.

zesterer: Ah, that's unfortunate :(

zesterer: Hmm, not sure I understood this message. So this does work?

stefnotch: Yep, this pull request should work.

zesterer:
Thanks for the PR!

stefnotch:
I'm assuming that the new version on crates.io includes this PR?

The links seem to be broken now, they point at `https://github.com/zesterer/chumsky/blob/VERGEN_IDEMPOTENT_OUTPUT/examples/brainfuck.rs`, and I can't quite tell why.

https://docs.rs/chumsky/1.0.0-alpha.5/chumsky/index.html#example-brainfuck-parser

stefnotch:
@zesterer I see. So was the vergen approach always doomed to fail on docs.rs? Or does that call for some tricky investigation?

PR #538, opened 2023-10-08
First attempt at filter labels

zesterer:


CraftSpider:
I'm for this personally - I avoid using `filter` in any of my projects largely because of this issue, always using `try_map` instead so I can mark my errors.

PR #544, opened 2023-10-10
fix: remove `debug_assertions` check causing `release` builds to fail

ErichDonGubler:


zesterer:
Thanks! I managed to miss this yesterday.

PR #552, opened 2023-10-23
Add `state` to pratt parser

ojkelly:
I recent updated, and needed to adjust a bunch of code for #537. I think it makes sense to reduce those to one `map_with`, however spending a few hours on it, I couldn't quite work out how to achieve the same here.

I tried passing a `MapExtra` instead of `span` and `state`. I got stuck on lifetime issues from nested closures, and loosing type inference on what `state` is.

Perhaps there's a way to make it work, but in the mean time this has unblocked me.

### DIFF
--- a/src/pratt.rs
+++ b/src/pratt.rs
@@ -104,13 +104,13 @@ where
 
     fn op_parser(&self) -> &Self::OpParser;
     fn associativity(&self) -> Associativity;
-    fn fold_infix(&self, _lhs: O, _op: Self::Op, _rhs: O, _span: I::Span) -> O {
+    fn fold_infix(&self, _lhs: O, _op: Self::Op, _rhs: O, _span: I::Span, _state: &mut <E as ParserExtra<'a, I>>::State) -> O {
         unreachable!()
     }
-    fn fold_prefix(&self, _op: Self::Op, _rhs: O, _span: I::Span) -> O {
+    fn fold_prefix(&self, _op: Self::Op, _rhs: O, _span: I::Span, _state: &mut <E as ParserExtra<'a, I>>::State) -> O {
         unreachable!()
     }
-    fn fold_postfix(&self, _lhs: O, _op: Self::Op, _span: I::Span) -> O {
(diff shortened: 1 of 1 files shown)


zesterer:
Thanks for the contribution! `MapExtra` is quite a new addition (as in: last week) so the pratt combinator has not been updated. I think this contribution is definitely *behaviourally* correct, but I think I'll hold off on merging it pending an attempt to make the pratt combinator support `MapExtra`. Thanks!

ojkelly:
> but I think I'll hold off on merging it pending an attempt to make the pratt combinator support `MapExtra`. Thanks!

Sounds good, if you can get `MapExtra` working that would be better.

zesterer:
I've added support in #553, hopefully this works for your use-case!

ojkelly:
Thanks! So that PR is what I had tried, but as it turns out I had somewhere else unintentionally constraint the type returned by a parsing function.

After wrangling all those, it's working.

I do still need to type it, but after rewriting the outer fn sig to get access to the Input and Extra generics inside the function it's working.

```rust
    expr.pratt((
      pratt::prefix(
        2,
        lexer::control::spanned_sub(),
        |op_token, rhs, e: &mut Extra<'tokens, '_, I, E>| {
          let span: Span = e.span();
          let state: &mut State = e.state();
          state.insert(crate::Node::ExpressionInlineOpNegative(
            Negative {
              span,
              op_token,
              rhs,
            },
          ))
        },
      ),
```

zesterer:
Yes, the need to add a type annotation is an unfortunate issue. It's a product of `rustc`'s weak inference algorithm that requires types to be fully known when calling methods on them or performing field accesses.

I say `rustc` and not 'Rust' more generally because `mrustc`, the Rust compiler written in C++, actually has a stronger constraint-based inference algorithm that *can* deal with cases like this.

If you want a neater way to write this inline, you can do

```rs
(e.state() as &mut State).insert(...);
```

Also, `span` shouldn't need an annotation here since no methods are being called on it.

ojkelly:
Ooh, I didn't think of doing that for the inline syntax, that is nice.

Yep dropping the span type works as well, I thought I needed it because it's a custom type, but it's all inferred well enough.

Issue #555, opened 2023-10-23
failed version of Ahash


GPeaky:
I runned a cargo update in my project as i usually do and i get this error:

error: failed to select a version for the requirement `ahash = "^0.7.0"`
candidate versions found which didn't match: 0.8.5, 0.8.4, 0.4.1, ...
location searched: crates.io index
required by package `hashbrown v0.12.3`
    ... which satisfies dependency `hashbrown = "^0.12.3"` of package `chumsky v0.9.0`
    ... which satisfies dependency `chumsky = "^0.9"` of package `lettre v0.11.0`
    ... which satisfies dependency `lettre = "^0.11.0"` of package `intelli v0.1.5 (C:\Users\Gerar\OneDrive\Documents\GitHub\intelli)`
perhaps a crate was updated and forgotten to be re-vendored?

The package that uses chumsky in my project is lettre and i checked if the last release version had updated ahash but i seams like is not the case

GPeaky:
I opened a pull request fixing this error:

https://github.com/zesterer/chumsky/pull/556

PR #556, opened 2023-10-23
Solution to fix ahash version error

GPeaky:


GPeaky:
Sorry i didn't put any information here. all the information is in this issue:
https://github.com/zesterer/chumsky/issues/555

sorry if i do something bad in github, i still learning about how to contribute 😞

zesterer:
Thanks!

zesterer:
Are you looking for a new version with this?

FWIW, I would recommend moving to one of the 1.0 alpha releases for any new work.

wackbyte:
In #555 they use the `lettre` crate which depends on `^0.9.0` :)

GPeaky:
> Are you looking for a new version with this?
> 
> 
> 
> FWIW, I would recommend moving to one of the 1.0 alpha releases for any new work.

Yeah am looking for 9.3 to open an issue in Letre to use 9.3 instead of 9.0 to fix this error cause now I cannot install lettre.

It's the alpha release totally compatible with 0.9?, and i probably think they will prefer to still using a release version

paolobarbolini:
> Are you looking for a new version with this?
> 
> FWIW, I would recommend moving to one of the 1.0 alpha releases for any new work.

We're looking into updating but there's no changelog for the 1.0 alpha releases and commits aren't tagged so it'll take a while. If you can backport the hashbrown upgrade I'd suggest doing so.

zesterer:
> It's the alpha release totally compatible with 0.9?

No, not at all. `1.0` is an almost from-scratch rewrite of the crate and some of the fundamental concepts (such as the way inputs have handled) have changes to allow for things like nested inputs, zero-copy parsing, etc. You might find [the guide](https://docs.rs/chumsky/1.0.0-alpha.6/chumsky/guide/index.html) to be useful. It's worth making the jump though: `1.0` is substantially faster and far more capable in innumerable ways.

> We're looking into updating but there's no changelog for the 1.0 alpha releases and commits aren't tagged so it'll take a while.

I meant a new `0.9.6` release. The `1.0` released (and the `main` branch) don't have this issue so should be unaffected.

GPeaky:
Perfet, if you could update this to fix this patch error

zesterer:
`0.9.3` has now been released.

GPeaky:
That's perfect, i already try it and its working now!, thanks

PR #568, opened 2023-11-16
Branch name updated to main in tutorial.md

morzel85:
One visible text and two links were still using the _master_ branch name but it looks like the _main_ is the right name...


zesterer:
Thanks. I'll merge this for now, although I'm planning to rewrite the tutorial entirely for 1.0.

PR #578, opened 2023-11-27
Fix benches and examples and check them in CI

### LINKED ISSUE #579: Failed to compile benches: no method named `slice` found
Hi there 👋 

I am trying to run benchmarks and getting bunch of error regarding missing method names:
```
$ cargo bench
...
error[E0599]: no method named `slice` found for struct `Repeated` in the current scope
   --> benches/lex.rs:129:53
    |
129 |         let digits = one_of(b'0'..=b'9').repeated().slice();
    |                                                     ^^^^^ method not found in `Repeated<OneOf<RangeInclusive<u8>, _, _>, _, _, _>`
...
```

Is there a way to get it working? I haven't really looked at the code precisely trying to fix the error, I had hoped that someone with knowledge of the library can give some tips.

Thanks!
Oleg

wackbyte:
This should prevent them from becoming out-of-date. I chose not to test them with the MSRV since it would be an unnecessary restriction. Closes #579.

Note: Is the use/position of `#[allow(clippy::let_and_return)]` okay, should it be moved, or should it be removed and fixed?

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -23,11 +23,11 @@ jobs:
             toolchain: nightly
             components: rustfmt, clippy
       - name: Run cargo check (all features)
-        run: cargo check --tests --verbose --all-features
+        run: cargo check --benches --examples --tests --verbose --all-features
       - name: Run cargo check (no features)
-        run: cargo check --tests --verbose --no-default-features
+        run: cargo check --benches --examples --tests --verbose --no-default-features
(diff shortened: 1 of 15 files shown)


wackbyte:
I bumped winnow to the latest version--should `cargo update` be run at some point to bump all dependencies since the `Cargo.lock` is checked in?

zesterer:
Thanks! All looks good here.

PR #592, opened 2024-01-27
Use diagnostic namespace, make message appear on ParserSealed mismatches

CraftSpider:
This should slightly improve the situation for trait errors. We don't get *too* much control over the diagnostic, but I tried to push the bits important to the front.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -3501,8 +3501,3 @@ mod tests {
         }
     }
 }
-
-fn foo() {
-    use prelude::*;
-    let p = any::<&str, extra::Default>().or(any().to(()));
-}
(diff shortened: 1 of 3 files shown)


CraftSpider:
Note that this will be an MSRV bump when it happens, but that the on_unimplemented diagnostic is *currently* in FCP to stabilize. That isn't a guarantee obviously, but it's a lot more likely to stabilize than a `rustc_attr`, which never will.

zesterer:
Awesome!

zesterer (approved):
You have merge perms, right?

CraftSpider:
I do - I'll merge this, then probably make another thing to fix the check build.

PR #595, opened 2024-02-01
Revert MapExtra lifetime commit (and re-move to input module)

zyansheep:


zyansheep:
Note to self: DONT FORGET TO RUN WITH ALL FEATURES ENABLED :sob:

zesterer:
Looks like there a few more things that need fixing as per the CI failures.

zyansheep:
Okay, it should be good to merge now. The only thing that might be of concern is the revert changed the slice() function from:
        `self.inp.slice(self.before..self.after)`
to:
        `self.inp.slice_since(self.before..)`

zesterer:
> Okay, it should be good to merge now. The only thing that might be of concern is the revert changed the slice() function from: `self.inp.slice(self.before..self.after)` to: `self.inp.slice_since(self.before..)`

That would be an incorrect change, I think `.slice(...)` is correct here. Any chance you could remake that change?

zyansheep:
Fixed the issue but I think latest nightly broke CI :(
Works locally on cargo 1.78.0-nightly (cdf84b69d 2024-02-02), but is broken in the CI on 2024-02-05.
My nix rust overlay isn't recent enough to test that version tho...

zesterer:
Argh, that's annoying, looks like we might need to update `ahash`. I'll merge this regardless, thanks!

zesterer:
Hmm, I admit I'm still a little confused by the original issue you had. What lifetime error were you running into again?

zyansheep:
I'm not sure exactly, but I think it may have had something to do with a lifetime being joined / not stored on the MapExtra struct.
The commit where it bugs is here: https://github.com/libdither/disp/tree/c65ac0d3e7e4f610bd5f3f18cf4a86d042dc5b6b and I bisected it back to the commit where the extra lifetime is removed.

PR #600, opened 2024-02-06
Added Non-Associative associativity for Pratt Parsers.

YunkaiZhang233:
In current Pratt parsers, only associativity type of `left` and `right` are supported. However, according to the original designs of Pratt parsing, it should also support Non-Associative infix binary operators.

For example, consider operators `==` as a non-associative operator. We should be able to identify it as non-associative and therefore rejecting `1 == 2 == 3` as an expression with the numbers being atomic literals. Whilst the current design only allow `==` to be either left-associative or right-associative.

This PR added associativity `non` for infix binary operators, and changed the binding power calculation function slightly based on [The Original Parsing Methodology](https://dl.acm.org/doi/pdf/10.1145/512927.512931) and [this article](https://www.engr.mun.ca/~theo/Misc/pratt_parsing.htm).

### DIFF
--- a/src/pratt.rs
+++ b/src/pratt.rs
@@ -10,6 +10,8 @@
 //! ['binding power'](https://matklad.github.io/2020/04/13/simple-but-powerful-pratt-parsing.html#From-Precedence-to-Binding-Power)
 //! that determines how strongly operators should bind to the operands around them.
 //!
+//! Here is ['another approach'](https://www.engr.mun.ca/~theo/Misc/pratt_parsing.htm) that provides a more comprehensive perspective of handling multiple operators and patterns
+//!
 //! Pratt parsers are defined with the [`Parser::pratt`] method.
 //!
 //! When writing pratt parsers, it is necessary to first define an 'atomic' operand used by the parser for building up
@@ -757,10 +759,9 @@ mod tests {
             ))
             .map(|x| x.to_string());
 
-        assert_eq!(
-            parser.parse("1+2!$*3").into_result(),
-            Ok("(((1 + (2!))$) * 3)".to_string()),
-        )
+        assert!(
+            parser.parse("§1+-~2!$*3").has_errors(),
+        );
     }
 
     #[test]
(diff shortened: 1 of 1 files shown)


src/pratt.rs
```diff
@@ -149,18 +153,33 @@ pub fn right(binding_power: u16) -> Associativity {
...
+            Self::Right(x) => *x as u32 * 3,
+            Self::Non(x) => *x as u32 * 3 + 1,
+        }
+    }
+
+    fn next_power(&self) -> u32 {
+        match self {
+            Self::Left(x) => *x as u32 * 3,
+            Self::Right(x) => *x as u32 * 3,
+            Self::Non(x) => *x as u32 * 3 - 1,
```
zesterer, line 182: For this to be valid, all powers should have 1 added to them such that `*x as u32 * 3 - 1` can never result in underflow.

src/pratt.rs
```diff
@@ -738,10 +759,9 @@ mod tests {
             ))
             .map(|x| x.to_string());
 
-        assert_eq!(
-            parser.parse("1+2!$*3").into_result(),
-            Ok("(((1 + (2!))$) * 3)".to_string()),
-        )
+        assert!(
+            parser.parse("§1+-~2!$*3").has_errors(),
+        );
```
zesterer, line 764: This change looks a little suspicious. Does this addition change the behaviour of existing parsers?

zesterer (commented):
Thanks for the PR, great work! Just a few comments.

src/pratt.rs
TheOnlyTails: Would this resolve the issue here?
```
            Self::Left(x) => *x as u32 * 3 + 1,
            Self::Right(x) => *x as u32 * 3 + 1,
            Self::Non(x) => *x as u32 * 3 - 1,
```

NomAnor:
Any progress?
 I stumbled on the same problem for == and was about to create an issue when I found this PR.

zesterer:
> Any progress? I stumbled on the same problem for == and was about to create an issue when I found this PR.

Another PR has been opened, but discussion is stale (and probably needs rebasing on upstream). If you're interested in picking it up again, I'd really appreciate it!

PR #601, opened 2024-02-08
docs: remove `ascii` namespace from `text`

magnusrodseth:
- When going through the tutorial, I noticed that the docs were somewhat outdated. 
- Updated imports related to `text` to ensure they compile


zesterer:
Thanks. FWIW the tutorial does need rewriting entirely for 1.0, but in the meantime this is a welcome addition.

zesterer:
Hmm, I think that this is a regression. The 1.0 release (which `main` is on) *adds* the `ascii` module. You're likely using the 0.9 release, which does not have this module.

magnusrodseth:
@zesterer My bad! That's OK 😊

PR #604, opened 2024-02-19
fix(examples/foo): Add to Cargo.toml and .expect() with usage

sshine:
Let me know if I understood the current state of things right.


zesterer:
Thanks!

PR #606, opened 2024-02-25
docs: hyperlink `then*` methods with each other

janriemer:
This hyperlinks `then`, `then_ignore` and `ignore_then` methods to each other and provides a short note when the other method(s) should be used. This improves discoverability of the methods and makes the similarity between those methods more obvious.

This is especially useful, when one navigates to those methods by using the _alphabetically sorted_ docs.rs menu on the side, because e.g. the methods `then_ignore` and `ignore_then` are pretty far apart in this side-menu despite their similar purpose.


zesterer:
Thanks! Seems like a good idea.

PR #626, opened 2024-04-26
Allow leading underscores in idents

calder:
# Summary

`_` is not a member of `XID_Start` because some languages don't consider it a valid leading character (certain SQL dialects for example). In practice, leading `_`'s are accepted by most languages and even have a special role in many modern languages where they indicate intentionally ignored variables. Accepting them in `ident()` should make it more useful and less surprising.

# References

* https://www.unicode.org/reports/tr31/
* https://www.unicode.org/reports/tr31/#NFKC_Modifications

### DIFF
--- a/src/text.rs
+++ b/src/text.rs
@@ -80,7 +80,7 @@ impl Char for char {
     }
 
     fn is_ident_start(&self) -> bool {
-        unicode_ident::is_xid_start(*self)
+        unicode_ident::is_xid_start(*self) || *self == '_'
     }
 
     fn is_ident_continue(&self) -> bool {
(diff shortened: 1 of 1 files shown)


src/text.rs
```diff
@@ -528,6 +528,33 @@ pub mod unicode {
...
+        fn test_err(self: Self, input: &'a str) {
+            assert_eq!(
+                self.parse(input),
+                ParseResult {
+                    output: None,
+                    errs: vec![EmptyErr::default()]
+                }
+            );
+        }
+    }
```
calder, line 556: Let me know if:
* This trait seems like overkill and you'd rather make these free functions.
* This trait seems generally useful and you'd rather move them to a `cfg(test)` block in `util.rs`.

zesterer: Hmm, I do think this is overkill. Free functions would be a little simpler.

zesterer (changes requested):
Thanks so much for the PR! I actually noticed this just a few days ago when playing around on a personal project, but didn't stop to check whether XID_Start actually permitted underscores.

src/text.rs
calder: Done.

zesterer:
Thanks again!

PR #632, opened 2024-05-04
Use diagnostic namespace on stable

wackbyte:
`#[diagnostic::on_unimplemented]` was stabilized in Rust 1.78.0 :)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,7 +23,7 @@ std = [
     "serde?/std"
 ]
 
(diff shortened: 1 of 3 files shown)


wackbyte:
(Should the MSRV be increased?)

zesterer:
Interesting question. Given that 1.0 is pre-release, I'd be inclined to say 'yes'. Obviously, when 1.0 releases, it'll be a good idea to pin it down a lot more.

PR #644, opened 2024-06-18
Update `Cargo.lock`

samestep:
As of commit ef9f20f6177b938550d0c12889c5d2dbac85558f, cloning and building the repo results in an extraneous diff on `Cargo.lock`, which this PR fixes.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -193,7 +193,7 @@ checksum = "baf1de4339761588bc0619e3cbc0120ee582ebb74b53b4efbf79117bd2da40fd"
 
 [[package]]
 name = "chumsky"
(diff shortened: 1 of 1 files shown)


baoyachi:
Why we manage Cargo.lock file in git
As a crates, there's no problem with ignore Cargo. lock

samestep:
@baoyachi I thought it was pretty well-accepted that lockfiles should be committed to version control, to make local development more deterministic? https://stackoverflow.com/a/44210813/5044950

But in any case, all I'm trying to do here is fix a tiny annoyance, nothing major.

zesterer:
This has since been resolved.

Issue #631, opened 2024-05-01
Labels don't show unless the EOI is in it's place


mnbjhu:
Suppose I have the following parser:

```rust
    let ident = select! {
        Token::Ident(ident) => ident,
    };
    just(Token::Keyword(Keyword::Select))
        .ignore_then(ident.labelled("column"))
        .then_ignore(just(Token::Keyword(Keyword::From)))
        .then(ident.labelled("table"))
        .map(|(column, from)| Statement::Select(Select { column, from }))
```
When I parse this input:

(Keyword(Select), 0..6)
(Ident("some"), 7..11)
(Keyword(From), 12..16)
(Newline, 16..17)

Then I get an error: 'found newline expected something else'. However if I use:

(Keyword(Select), 0..6)
(Ident("some"), 7..11)
(Keyword(From), 12..16)

Then I get the error: 'found end of input expected table' which is what I'd expect. Is this a bug or am I missing something? (1.0.0-alpha.7)

zesterer:
Hmm, this odd and seems unintended. Do you have a minimal working example of this I can play with?

mnbjhu:
Of course, I've created an example here: https://github.com/mnbjhu/label_test

ojkelly:
I noticed this after attempting to update, and found it only occurs on sources without a newline. I'm working on a project with a few hundred tests without newlines (unit tests for individual parsers for example).

I had a dig through the diff from what I had been using and found what I think might be the cause https://github.com/zesterer/chumsky/pull/645

I'm not fully across the details of that area of code, so it may not be the ideal fix but hopefully it's useful.

mnbjhu:
Thanks for this, I'll make sure to check it out this weekend. I've been working on a language server recently and this will be really useful :)

PR #645, opened 2024-06-22
fix: #631 revert change in label.rs to fix error labels

ojkelly:
When a source does not contain a newline,  labels are lost/ignored and replaced with the fallback `expected something else`.

This reverts the change made in this commit which appears to fix this https://github.com/zesterer/chumsky/commit/68375371a5fde6ee14f190c14e9a9cee0697f022#diff-3d2dfe7e70ee2398dd0941e8f6ea855652695c3f82b7816f649668fb2d56b93dR60

### DIFF
--- a/src/label.rs
+++ b/src/label.rs
@@ -57,7 +57,7 @@ where
         inp.errors.alt = old_alt;
 
         if let Some(mut new_alt) = new_alt {
-            let before_next = before.offset.into();
+            let before_next = before.offset.into() + 1;
             if new_alt.pos.into() == before_next {
                 new_alt.err.label_with(self.label.clone());
             } else if self.is_context && new_alt.pos.into() > before_next {


zesterer:
Hmm. I am a bit suspicious of this, but I need to take another look at it at some point. Do tests continue to pass with `--features label`?

ojkelly:
> Hmm. I am a bit suspicious of this, but I need to take another look at it at some point.

Yeah me too, I'm not familiar with this part of the codebase. I had a play around with the diffs to see what would fix it for my repo and this has.

Looks like the tests are failing.

xldenis:
I seem to be encountering this issue in `alpha.7` when using `select!` in combination with a logos lexer. I'll need to investigate further to minimize.

zesterer:
> I replicated this issue locally with a parser and confirmed this patch fixes it. Additionally, the CI failures seem to be spurious failures caused by `time`. I think if @zesterer restarts CI it should pass.

As a point of interest, do you have an example of what you think the correct behaviour is that you can turn into a test? That would be very useful for figuring out the best way to implement a fix.

PR #664, opened 2024-08-31
Add recursive_fold and recursive_2 to prevent memory leaks

jkylling:
By using `recursive_2` or `recursive_fold` instead of `Recursive::declare` we can prevent situations where we either keep too many strong references around, leading to memory leaks, or too few strong references around, leading to broken parsers because all the strong references and the parsers are dropped.

We could consider removing `Recursive::declare` from the public API, since it is hard to use it correctly.

Please see the investigation (for version 0.9) in https://github.com/zesterer/chumsky/issues/486#issuecomment-2198514803

If code is migrated to use `recursive_2` or `recursive_fold` we should be able to resolve https://github.com/zesterer/chumsky/issues/486 I have tested that this patch for 0.9 fixes the memory issue on jaq when repeatedly creating parsers.

### DIFF
--- a/src/recursive.rs
+++ b/src/recursive.rs
@@ -134,7 +134,7 @@ impl<'a, 'b, I: Input<'a>, O, E: ParserExtra<'a, I>> Recursive<Indirect<'a, 'b, 
     /// Defines the parser after declaring it, allowing it to be used for parsing.
     // INFO: Clone bound not actually needed, but good to be safe for future compat
     #[track_caller]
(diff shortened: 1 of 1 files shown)


zesterer:
How do you think this solution compares to #494 ?

jkylling:
> How do you think this solution compares to #494 ?

I have not looked at #494 in detail, but I believe it allows accidentally dropping one of the parsers while keeping the other one. That would render the remaining parser unusable. The API should try to prevent accidental misuse if possible.

PR #673, opened 2024-09-24
Update tutorial.md for v0.9.3

jnoortheen:
I was just trying out the tutorial and found that the functions are moved out.


zesterer:
Thanks, I guess this is an improvement on the existing state of affairs. Really, it needs updating for 1.0 at some point.

PR #676, opened 2024-09-28
Fix example code in Repeated::exactly

istathar:
There were seven Rings of Power given to the Dwarven Lords, not six.

> Three Rings for the Elven-kings under the sky,
> Seven for the Dwarf-lords in their halls of stone,
> Nine for Mortal Men doomed to die,
> One for the Dark Lord on his dark throne
> In the Land of Mordor where the Shadows lie.

Clearly this is a very serious bug in **chumsky** and making this fix will dramatically improve the correctness and performance of the library.

### DIFF
--- a/src/combinator.rs
+++ b/src/combinator.rs
@@ -1385,7 +1385,7 @@ where
     ///
     /// let for_the_dwarves = ring
     ///     .repeated()
-    ///     .exactly(6)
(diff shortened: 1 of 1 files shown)


zesterer:
Frankly, I'm shocked that this went unnoticed for so long. Thanks for the vital fix!

PR #678, opened 2024-10-11
Changed `Or` path choosing

### LINKED ISSUE #677: `or` choosing secondary option, over recovered first option
I'm building an lsp, so error recovery is very important for me. I wanted to recover a function call that would look something like that `foo(`

My parser would look like this:
```rs
function_call_parser()
  .or(var_parser())
  .or(parentheses_parser())
```

`function_call_parser` correctly recovers from this error and maps to proper value, but The final result would be a combination of a var `foo` with parentheses `()` showing an error form parentheses parser.

I've looked at the `Or` struct implementation and I think this is the problem:

```rs
if a_res.0.is_empty() {
    if let (a_errors, Ok(a_out)) = a_res {
        return (a_errors, Ok(a_out));
    }
}

stream.revert(pre_state);

#[allow(deprecated)]
let b_res = debugger.invoke(&self.1, stream);
let b_state = stream.save();

if b_res.0.is_empty() {
    if let (b_errors, Ok(b_out)) = b_res {
        return (b_errors, Ok(b_out));
    }
}
```

If I understand this correctly, if the second parser doesn't produce an error, while the first does, we choose the result of the second parser even tho the first parser may have parsed correctly.

While this may be a valid behaviour it's not consistent with what the docs say:
"If both parser succeed, the output of the first parser is guaranteed to be prioritised over the output of the second."

Is there some other util I can use, or should I create my own? I can create the fix as it seems pretty easy. What do you think?

KrosFire:
resolves #677 

The Or parser will now choose a path that succeeded first and if both fail it will try to figure out witch error is better suited

### DIFF
--- a/src/combinator.rs
+++ b/src/combinator.rs
@@ -38,13 +38,8 @@ impl<I: Clone, O, A: Parser<I, O, Error = E>, B: Parser<I, O, Error = E>, E: Err
         let a_res = debugger.invoke(&self.0, stream);
         let a_state = stream.save();
 
-        // If the first parser succeeded and produced no secondary errors, don't bother trying the second parser
-        // TODO: Perhaps we should *alwaus* take this route, even if recoverable errors did occur? Seems like an
-        // inconsistent application of PEG rules...
-        if a_res.0.is_empty() {
-            if let (a_errors, Ok(a_out)) = a_res {
-                return (a_errors, Ok(a_out));
-            }
+        if let (a_errors, Ok(a_out)) = a_res {
+            return (a_errors, Ok(a_out));
         }
 
         stream.revert(pre_state);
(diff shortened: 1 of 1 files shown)


zesterer:
Oh, this is for 0.9? Hmm...

I am somewhat less inclined to merge this because 0.9 is not maintained any longer, bar fixes for significant bugs. Is there a particular reason that you're using 0.9 instead of 1.0?

KrosFire:
The first version is still in alpha and is not the default in cargo so I was using 0.9.3 :D

If I understand correctly the current implementation of Or in the main branch it's just a syntax sugar for choice, which returns first result that parses with success, so there is no need to change anything.

Is the newest version stable enough and does it have all the functionalities to use instead of 0.9? I'm looking forward to it because of the performence and dropping the `Rc`

zesterer:
> Is the newest version stable enough and does it have all the functionalities to use instead of 0.9?

I'd say so, yes. There are little changes to come that are *technically* breaking, but only in the most abstract sense: Almost all parsers written against 1.0-alpha should compile or compile with only very minor changes come 1.0 stable. The design is maybe 85% there, and the features that exist now aren't going anywhere. I'm avoiding a stable release for now purely because there are some minor API burrs still to sand down.

Issue #96, opened 2022-03-03
Support integration with cstree


zesterer:
With the eventual merge of #82, it might be possible to have chumsky integrate with `cstree`, a library for lossless parsing using untyped syntax tree. This might be achieved by allowing implementers of the `Input` trait to specify functions that should be run when sequences of the input are consumed by the parser. For a dedicated `Input` implementation (that wraps another internally) we could specify these functions, allowing the emission of parse events.

zesterer:
This is probably possible with `zero-copy`, but it's not on the roadmap for release.

robo9k:
What does `zero-copy` refer to (a Git branch?), the mentioned pull request has been merged already?
Are we talking about `InputRef` (not `Input`) here and if so, how would that work given it's not a trait?
What would `chumsky`'s role be; parse input (possibly from a separate lexer) and output .. events to construct a `cstree` CST?

CraftSpider:
`zero-copy` used to be a git branch, but has since been merged and is now the main branch of the repository, anything after `1.0.0-alpha.1` is zero-copy.

Most likely this would involve both `Input` and `InputRef` - `InputRef` would gain methods that the parsers call, that it delegates to implementors of the `Input` trait (or possible a new sub-trait of it, similar to `SliceInput` or `ValueInput`, if I understand the original idea correctly.

I can't answer this one as zesterer, but personally, that sounds about right, given my limited knowledge of cstree.

zesterer:
Yep, that's about right: we'd expand the input traits to allow collecting 'parsing events'. For most imputs, this would just be a stub that does nothing, allowing the optimiser to swallow the overhead. For a cstree input, these parsing events could be consumed by cstree's node builder and used to track token inputs.

Unfortunately I don't have the time to put together a proof of concept for this at the moment, but it probably wouldn't be the most difficult thing in the world to do (for those that understand how cstree works well).

jyn514:
@zesterer helped me got something working that functions pretty well:
```rust
struct RowanNode_<'a, O, P: CSTParser<'a, O>> {
    parser: P,
    kind: SyntaxKind,
    _marker: PhantomData<(&'a str, fn() -> O)>,
}

type RowanNode<'a, O, P> = Ext<RowanNode_<'a, O, P>>;

/// This needs to be an extension, not a combinator using `map_with`, because map_with can be evaluated multiple times in the case of backtracking.
[...]

PR #681, opened 2024-10-18
Add a new `ParserExtra::State: Inspector` trait

jyn514:
This allows using imperative concrete syntax tree parsers like `rowan` and `cstree`. In particular those libraries want to know every token that is parsed, and need to know when chumsky backtracks and reparses the same tokens again.

This adds the following new API surface:
```rust
pub trait Inspector<'a, I: Input<'a>>: Default {
    type SaveMarker: Copy + Clone;
    fn on_token(&mut self, token: &I::Token);
    fn on_save<'parse>(&self, offset: I::Offset) -> Self::SaveMarker;
    fn on_rewind<'parse>(&mut self, marker: Marker<'a, 'parse, I, Self::SaveMarker>);
}
pub struct SimpleState<T>(pub T);
impl<'a, T, I: Input<'a>> Inspector<'a, I> for SimpleState<T>;
impl<T> DerefMut<Target = T> for SimpleState<T>;
impl<T> From<T> for SimpleState<T>;
```
and additionally now requires `ParserExtra::State: Inspector`.

i didn't write any of my own tests yet but my own parser that hooks chumsky up to cstree does work: https://github.com/spreadsheet-lang/spreadsheet/commit/1b9731311aa97c1130c76de67966052909c35f75#diff-14962dfdd56397d99e07041d6337cb5b0327f4d7433d80fd3383955415014910R111

Fixes https://github.com/zesterer/chumsky/issues/96


src/recorder.rs
```diff
@@ -0,0 +1,71 @@
...
+//! and rubber bungs in his ears. He knew exactly what was happening and had known
+//! ever since his Sub-Etha Sens-O-Matic had started winking in the dead of night
+//! beside his pillar and woken him with a start."*
+use crate::{input::Marker, Input};
+use core::ops::{Deref, DerefMut};
+
+#[allow(unused)] // for intra-doc links
+use crate::Parser;
+
+/// A type that receives event hooks when certain parsing actions occur.
```
zesterer, line 13: I think I'd be tempted to just call this type `State` and remove the existing type alias(es) (`State`, `Context`, and `Err`). `Recorder` seems too specific a name for what this might eventually turn into.

src/recorder.rs
```diff
@@ -0,0 +1,71 @@
...
+///
+/// If you don't need to receive event hooks, use [`SimpleState`].
+pub trait Recorder<'a, I: Input<'a>> {
+    /// A type the Recorder can use to revert to a previous state.
+    ///
+    /// For implementation reasons, this is required to be `Copy + Clone`.
+    type SaveMarker: Copy + Clone;
+
+    /// This function is called when a new token is read from the input stream.
+    fn on_token(&mut self, token: I::Token);
```
zesterer, line 25: Is this ever getting invoked? I assume not, since this sadly doesn't generalise to by-ref and by-value inputs. The correct thing would be `token: I::TokenMaybe`.

src/recorder.rs
```diff
@@ -0,0 +1,71 @@
...
+    fn deref_mut(&mut self) -> &mut Self::Target {
+        &mut self.0
+    }
+}
+
+impl<T> From<T> for SimpleState<T> {
+    fn from(value: T) -> Self {
+        Self(value)
+    }
+}
```
zesterer, line 71: Should probably also add `From<SimpleState<T>> for T` (does this break the orphan rules nowadays, or will it need to be `Into`? I can't remember).

zesterer (changes requested):

src/recorder.rs
zesterer: FWIW, the place to call it is in these functions:

- `InputRef::skip_while`
- `InputRef::next_inner`
- `InputRef::next_maybe_inner`
- `InputRef::next_ref_inner`

src/recorder.rs
jyn514: even Into breaks orphan rules unfortunately, because we could have `T: From<SimpleState>`:
```
 2  error[E0119]: conflicting implementations of trait `Into<_>` for type `SimpleState`
 e<_>`
   --> src/recorder.rs:72:1
    |
 72 | impl<T> Into<T> for SimpleState<T> {
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: conflicting implementation in crate `core`:
            - impl<T, U> Into<U> for T
              where U: From<T>;
```
i think `state.0` is not too bad to write though.

src/recorder.rs
jyn514: i'm ok with renaming but "State" sounds too generic, i was hoping we could indicate that this is receiving events somehow. i've renamed it to `Inspector` for now.

src/recorder.rs
jyn514: it was really a pain to convert back and forth between `MaybeRef` :sweat_smile: i added those calls and changed the signature to pass in `&Token` instead.

zesterer:
Looks like there are a few minor CI failures

zesterer:
Thanks so much for your work on this, it's very appreciated!

PR #682, opened 2024-10-20
Added support for pratt parsers with up to 676 operators

### LINKED ISSUE #658: Pratt parser only implemented for tuples A..Z
I really appreciate the pratt parser feature. However, it is currently limited in the number of operators you can implement with it, since `impl_pratt_for_tuple!(A_ B_ C_ D_ E_ F_ G_ H_ I_ J_ K_ L_ M_ N_ O_ P_ Q_ R_ S_ T_ U_ V_ W_ X_ Y_ Z_);` only supports up to 26 operations. 
Maybe this can be fixed by using a proc macro to implement the Pratt parser for tuples of size N, where N can be supplied in some way by the library user? I'm no expert on macros though. A quick fix that would suffice for most cases would probably be to just double the tuple length?

TheOnlyTails:
Adds support for 676-operator pratt parsers (AA..=ZZ)
(this took exactly 5 minutes using multi-cursors)

Closes #658


zesterer:
Hmm, I think this is likely to have a non-trivial impact on compilation time. Remember that the macro system still needs to generate all of those impls and the type system needs to check them, even if they go unused. 

I'm happy to see the number of operators supported by `pratt` increased a little: let's say, to 64. But I think this is excessive.

TheOnlyTails:
Agreed, I tested it out and it crashed my editor when running clippy, not to mention having to increase the recursion limit for macro expansion.

wackbyte:
Maybe it could be supported with nested tuples? but that might be too hacky

zesterer:
It might be a possibility, yes.

TheOnlyTails:
I'll see if I can make that a reality.

src/pratt.rs
```diff
@@ -513,16 +513,132 @@ macro_rules! impl_pratt_for_tuple {
...
-    GA_ GB_ GC_ GD_ GE_ GF_ GG_ GH_ GI_ GJ_ GK_ GL_ GM_ GN_ GO_ GP_ GQ_ GR_ GS_ GT_ GU_ GV_ GW_ GX_ GY_ GZ_
-    HA_ HB_ HC_ HD_ HE_ HF_ HG_ HH_ HI_ HJ_ HK_ HL_ HM_ HN_ HO_ HP_ HQ_ HR_ HS_ HT_ HU_ HV_ HW_ HX_ HY_ HZ_
+    O1 O2 O3 O4 O5 O6 O7 O8
+    O9 O10 O11 O12 O13 O14 O15 O16
+    O17 O18 O19 O20 O21 O22 O23 O24
+    O25 O26
 );
 
+#[allow(unused_variables, non_snake_case)]
+impl<'a, Atom, T, const N: usize> Pratt<Atom, [T; N]> {
```
zesterer, line 523: I don't think this is doing what you think it's doing, unfortunately. This requires that all operators are of exactly the same time, which is almost useless. For this to be useful you'd need to implement `Operator` for `Box<dyn Operator>`, but then the code below like `T::IS_PREFIX` and others would need to change to check what kind of operator it is at run time.

zesterer:
Closing this since #685 added support for nested pratt operators.

PR #683, opened 2024-10-21
Overhauled Input trait to improve versatility, added IterInput

zesterer:
This PR does a few things:

- Improves the design of `Input`, allowing is to support a wider array of input types (see the new `IterInput`)
- Gives inputs mutable access to part of their cached state, removing the need for `Cell`/`RefCell` in existing input implementations
- Removes the `Sealed` bound from all input types except `StrInput`, allowing users to implement their own input types

Review/thoughts welcome!

If you don't want to look at the whole diff, the important part is the new definition of `Input`.

The performance implications are unclear: Some benchmarks are slightly faster, some are slightly slower. I don't think it's notable enough to matter.


zesterer:
The more I think about this, the more I'm confident that this MR is broadly the right thing to do, so I'm going to merge it to avoid unnecessary rebasing work. Post-merge review is still welcome and useful though!

Issue #638, opened 2024-05-19
docs.rs links to GH files appear broken


inferiorhumanorgans:
For the alpha7 pages at least:

e.g. On this section:

https://docs.rs/chumsky/1.0.0-alpha.7/chumsky/index.html#example-brainfuck-parser

The link to `brainfuck.rs` is: `https://github.com/zesterer/chumsky/blob/VERGEN_IDEMPOTENT_OUTPUT/examples/brainfuck.rs`

inferiorhumanorgans:
Quick addendum, this seems confusing:

https://docs.rs/chumsky/1.0.0-alpha.7/chumsky/trait.Parser.html#method.ignore_with_ctx

"If you don’t need the context in the output, use Parser::then_with_ctx."

https://docs.rs/chumsky/1.0.0-alpha.7/chumsky/trait.Parser.html#method.then_with_ctx

"If you don’t need the context in the output, prefer Parser::ignore_with_ctx."

PR #687, opened 2024-10-22
then_docs: fix what was likely a copy-paste error

Zij-IT:
This addresses the addendum within #638, but notably does _not_ fix the primary part of the issue, as that will have to wait until the next release.


zesterer:
Thanks!

PR #690, opened 2024-10-26
Add input traits implementation for `Graphemes`

Hedgehogo:
This is to prevent constructs like this `̲0` from being treated as a symbol followed by an number. 

P.S. Extended grapheme clusters are being used.

### DIFF
--- a/src/input.rs
+++ b/src/input.rs
@@ -12,6 +12,7 @@ pub use crate::stream::{BoxedExactSizeStream, BoxedStream, IterInput, Stream};
 use super::*;
 #[cfg(feature = "std")]
 use std::io::{BufReader, Read, Seek};
+use unicode_segmentation::{Graphemes, UnicodeSegmentation};
 
 /// A trait for types that represents a stream of input tokens. Unlike [`Iterator`], this type
 /// supports backtracking and a few other features required by the crate.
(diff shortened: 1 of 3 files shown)


src/input.rs
```diff
@@ -283,6 +284,84 @@ impl<'src> SliceInput<'src> for &'src str {
...
+        cursor: &mut Self::Cursor,
+    ) -> Option<Self::MaybeToken> {
+        if *cursor < this.len() {
+            // SAFETY: `cursor < self.len()` above guarantees cursor is in-bounds
+            //         We only ever return cursors that are at a character boundary
+            let c = this
+                .get_unchecked(*cursor..)
+                .graphemes(true)
+                .next()
+                .unwrap_unchecked();
```
zesterer, line 318: I think the safety of this unwrap should be justified too: How do we know that simply because `cursor < this.len()` there will always be at least one grapheme produced? Is every non-empty run of codepoints *guaranteed* to produce a grapheme?

src/input.rs
```diff
@@ -283,6 +284,84 @@ impl<'src> SliceInput<'src> for &'src str {
...
+
+impl<'src> ValueInput<'src> for Graphemes<'src> {
+    #[inline(always)]
+    unsafe fn next(this: &mut Self::Cache, cursor: &mut Self::Cursor) -> Option<Self::Token> {
+        Self::next_maybe(this, cursor)
+    }
+}
+
+impl<'src> SliceInput<'src> for Graphemes<'src> {
+    type Slice = Graphemes<'src>;
```
zesterer, line 347: FWIW using `type Slice = Graphemes<'src>` means that this input will be incompatible with, say, the regex parser. I assume that's fine given that regex is a blunt tool that might cause a problem for grapheme boundaries, but it's worth considering.

zesterer (commented):
Thanks, I broadly agree that this is a good thing to merge. Maybe it could be moved to `text.rs` though?

src/input.rs
Hedgehogo: I can say based on the `next` implementation that `None` is only returned when the end of the string is encountered. 

```rs
fn next(&mut self) -> Option<&'a str> {
    let start = self.cursor.cur_cursor();
    if start == self.cursor_back.cur_cursor() {
        return None;
    }
    let next = self.cursor.next_boundary(self.string, 0).unwrap().unwrap();
    Some(&self.string[start..next])
}
```

But you can't rely on the implementation, so let's look at [standard](https://unicode.org/reports/tr29/#Grapheme_Cluster_Boundary_Rules). The first clause says that the grapheme cluster is terminated at the end of the string. It is not said that this situation is wrong.

There is also this not very clear paragraph referring to such a guarantee: 
> The continuing characters include nonspacing marks, the Join_Controls (U+200C ZERO WIDTH NON-JOINER and U+200D ZERO WIDTH JOINER) used in Indic languages, and a few spacing combining marks to ensure canonical equivalence. There are cases in Bangla, Khmer, Malayalam, and Odiya in which a ZWNJ occurs after a consonant and before a virama or other combining mark. These cases should not provide an opportunity for a grapheme cluster break. Therefore, ZWNJ has been included in the Extend class. **Additional cases need to be added for completeness, so that any string of text can be divided up into a sequence of grapheme clusters.** Some of these may be degenerate cases, such as a control code, or an isolated combining mark.

zesterer: Fair enough. Could these two points be added as justification in a `// SAFETY:` comment?

Hedgehogo: > Fair enough. Could these two points be added as justification in a `// SAFETY:` comment?

What exactly did you call two points? There are 3 separate arguments here.

zesterer: In particular, that the existing implementation doesn't allow it to fail and also that there's justification in the unicode standard for it not failing in the future either.

zesterer:
Thanks for the PR!

PR #718, opened 2025-01-14
Forward NONCONSUMPTION_IS_OK in IterParser implementations

jojastahl:
IterParser Combinators that depend on other IterParsers need to forward their NONCONSUMPTION_IS_OK flag.

### DIFF
--- a/src/combinator.rs
+++ b/src/combinator.rs
@@ -92,6 +92,8 @@ where
     where
         I: 'src;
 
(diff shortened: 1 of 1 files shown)


zesterer:
Thanks, nice catch!

zesterer:
Looks like it needs a `cargo fmt`.

jojastahl:
cargo fmt done and added a fix

PR #720, opened 2025-01-26
Fix some holes in documentation

benjamin-cates:
Hello!

Added some missing documentation, mostly examples and filling in holes in the error module.

Change list:
- Added examples to error::EmptyError, error::Cheap, error::Simple, Error::Rich
- Added examples to Parser::into_iter, Parser::to_slice
- Added note comparing IterParser::Collect and Parser::to_slice on including/excluding ignored patterns.
- Added links to relevant attribute types in errors
  - It wasn't really clear what error::Rich was storing, especially that the links to RichPattern and RichReason were buried in method headers below, so I thought it would be good to add links to those.
  - It also wasn't clear what the return type of each of the "span" methods was, since they all said they returned "&S", but for some reason, cargo doc linked to [https://doc.rust-lang.org/std/primitive.reference.html](https://doc.rust-lang.org/std/primitive.reference.html) :sob:
- Fixed inconsistency between docs on Simple::found and Rich::found

Let me know if I should make any changes

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2036,7 +2036,21 @@ pub trait Parser<'src, I: Input<'src>, O, E: ParserExtra<'src, I> = extra::Defau
     /// The resulting iterable parser will emit each element of the output type in turn.
     ///
     /// This is *broadly* analogous to functions like [`Vec::into_iter`], but operating at the level of parser outputs.
-    // TODO: Example
+    ///
+    /// # Examples
+    /// 
+    /// ```
+    /// use chumsky::prelude::*;
+    /// let letter = one_of::<_, _, extra::Err<Simple<char>>>("0123456789");
+    /// // Match a four digit number
+    /// let list = letter
+    ///     .repeated()
+    ///     .to_slice()
+    ///     .map(|string: &str| string.chars())
+    ///     .into_iter()
+    ///     .collect_exactly::<[char; 4]>();
+    /// assert_eq!(list.parse("1235").unwrap(), ['1', '2', '3', '5']);
+    /// ```
     fn into_iter(self) -> IntoIter<Self, O>
     where
         Self: Sized,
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -2036,7 +2064,21 @@ pub trait Parser<'src, I: Input<'src>, O, E: ParserExtra<'src, I> = extra::Defau
...
+    /// ```
+    /// use chumsky::prelude::*;
+    /// let letter = one_of::<_, _, extra::Err<Simple<char>>>("0123456789");
+    /// // Match a four digit number
+    /// let list = letter
+    ///     .repeated()
+    ///     .to_slice()
+    ///     .map(|string: &str| string.chars())
+    ///     .into_iter()
+    ///     .collect_exactly::<[char; 4]>();
```
zesterer, line 2079: I feel like this is a somewhat contrived example since the code could be expressed equivalently but more simply like so:

```rs
let list = letter.repeated().collect_exactly::<[char; 4]>();
```

A better example might be something like:

```rs
// Parses whole integers
let num = text::int().padded().map(|x| x.parse::<u64>().unwrap());

// Parses a range like `0..4` into a vector like `[0, 1, 2, 3]`
let range = num.ignore_then(just("..")).then(num)
    .map(|(x, y)| x..y)
    .into_iter()
    .collect::<Vec<_>>();
// Parses a list of numbers into a vector
let list = num.separated_by(just(',')).collect::<Vec<_>>();

assert_eq!(list.parse("0, 1, 2, 3").unwrap(), [0, 1, 2, 3]);
assert_eq!(list.parse("0..4").unwrap(), [0, 1, 2, 3]);
```

zesterer (changes requested):
Thanks for the PR! I think these changes are great. I've just one suggestion regarding one of the examples.

src/lib.rs
benjamin-cates: Oh okay, thank you! I was struggling to think of a useful example lol

zesterer: It's difficult! A lot of chumsky's parsers only become useful at scale, so coming up with short examples that demonstrate their use it tough.

zesterer:
Thanks for the PR!

Issue #719, opened 2025-01-26
SimpleSpan::new method name conflict


benjamin-cates:
In the impl block for `SimpleSpan`, there is a method `new` defined as follows:
```rust
impl<T> SimpleSpan<T> {
    pub fn new(start: T, end: T) -> SimpleSpan<T> {
        SimpleSpan {
            start,
            end,
            context: (),
        }
    }
}
```

However there is also the definition for a method of the same name in the `Span` trait.
```
impl<T: Clone, C: Clone> Span for SimpleSpan<T, C> {
    type Context = C;
    type Offset = T;

    fn new(context: Self::Context, range: Range<Self::Offset>) -> Self {
        Self {
            start: range.start,
            end: range.end,
            context,
        }
    }
}
```

Since they have the same name, to call Span::new, you would have to write `<SimpleSpan as Span>::new`, which is... not ideal.

I noticed this struct hasn't been published in a stable release yet, so I would like to suggest changing the method name.

If you have any suggestions, I would be happy to submit a PR ☺️

zesterer:
I think removing `SimpleSpan::new` in favour of `Span::new` is a good idea: the latter is more general. `SimpleSpan::new` originally existed because `SimpleSpan` didn't support contexts (i.e: it had a default context of `()`) but this has since changed.

PR #721, opened 2025-01-28
Fix SimpleSpan method headers

benjamin-cates:
Changelog
- Remove SimpleSpan::new, which can be replaced with <SimpleSpan as Span>::new
- Remove SimpleSpan::splat
- Update SimpleSpan into range or into iterator to support contexts

Fixes: #719

### DIFF
--- a/src/span.rs
+++ b/src/span.rs
@@ -94,15 +94,6 @@ pub struct SimpleSpan<T = usize, C = ()> {
     context: C,
 }
 
-impl<T: Clone, C: Clone> SimpleSpan<T, C> {
-    /// Create a new `SimpleSpan` from a single offset, useful for an EOI (End Of Input) span.
-    ///
-    /// Context is the unit type `()` by default.
-    pub fn splat(context: C, offset: T) -> SimpleSpan<T, C> {
-        Self::new(context, offset.clone()..offset)
-    }
-}
-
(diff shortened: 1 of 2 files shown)


src/span.rs
```diff
@@ -128,8 +111,8 @@ impl<T> From<Range<T>> for SimpleSpan<T> {
     }
 }
 
-impl<T> From<SimpleSpan<T>> for Range<T> {
-    fn from(span: SimpleSpan<T>) -> Self {
+impl<T, C> From<SimpleSpan<T, C>> for Range<T> {
```
zesterer, line 114: This impl worries me slightly. `From`/`To` are best used for lossless conversions, but adding support for a non-empty context means that this conversion is no longer lossless. The context is a meaningful aspect of a span, and I don't think hiding that is a good idea. For users that really do want to forget the context (for example, as the last step in diagnostic generation) there's the `SimpleSpan::to_range` method.

zesterer (changes requested):
Broadly happy with this PR, but I'm squirming a bit about the change to the `From` impl. Thanks for opening!

src/span.rs
benjamin-cates: Okie I'll remove that

zesterer:
Thanks for the PR!

PR #723, opened 2025-01-30
chore: Update to Winnow 0.7

epage:
The further changes to the bench are to bring it slightly closer to how our own json example looks which is more idiomatic for the needs of that example.


zesterer (approved):
Thanks!

PR #728, opened 2025-02-06
Attempt At Non-Infix Pratt Parsing

lancylot2004:
Along the same lines as #600.

Currently, the `chumsky::pratt` parser only supports binary infix operators with `left` or `right` associativity, but not non-infix ones. Other libraries and literature might term this as `InfixN` or non-associative. (Please correct me if I'm wrong~)

For example, comparison operators like `>` and `==` are typically non-associative, since expressions like `1 == 2 == 3` is invalid. (They cannot be chained.)

This PR adds a `non` option for associativity, and tries to follow the same [article](https://www.engr.mun.ca/~theo/Misc/pratt_parsing.htm).

### DIFF
--- a/src/pratt.rs
+++ b/src/pratt.rs
@@ -492,23 +492,21 @@ pub fn non(binding_power: u16) -> Associativity {
 
 impl Associativity {
     fn left_power(&self) -> u32 {
-        match self {
-            &Self::Left(x) | &Self::Non(x) => x as u32 * 3,
-            &Self::Right(x) => x as u32 * 3 + 1,
-        }
+        let (&Self::Left(x) | &Self::Non(x) | &Self::Right(x)) = self;
+        x as u32 * 3
     }
 
     fn right_power(&self) -> u32 {
         match self {
-            &Self::Right(x) | &Self::Non(x) => x as u32 * 3,
-            &Self::Left(x) => x as u32 * 3 + 1,
+            &Self::Left(x) | &Self::Non(x) => x as u32 * 3 + 1,
+            &Self::Right(x) => x as u32 * 3,
         }
     }
 
     fn next_power(&self) -> u32 {
         match self {
             &Self::Left(x) | &Self::Right(x) => x as u32 * 3,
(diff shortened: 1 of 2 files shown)


zesterer:
Thanks for the PR, should be able to take a look at this soon.

src/pratt.rs
```diff
@@ -1254,4 +1300,25 @@ mod tests {
...
+                // -- Infix
+                infix(left(1), just('+'), |l, _, r, _| i(Expr::Add, l, r)),
+                infix(non(2), just('*'), |l, _, r, _| i(Expr::Mul, l, r)),
+            ))
+            .map(|x| x.to_string());
+        assert_eq!(
+            parser.parse("1+2*3").into_result(),
+            Ok("(1 + (2 * 3))".to_string())
+        );
+        assert!(parser.parse("1+2*3*3").has_errors());
```
zesterer, line 1322: I think it would be a good idea to check that `1+2*3*5` fails here too. `assert!(parser.parse(...).has_errors())` should be sufficient.

lancylot2004: Added!

src/pratt.rs
```diff
@@ -462,18 +481,32 @@ pub fn right(binding_power: u16) -> Associativity {
...
+        x as u32 * 3 + 1
+    }
+
+    fn right_power(&self) -> u32 {
         match self {
-            Self::Left(x) => *x as u32 * 2,
-            Self::Right(x) => *x as u32 * 2 + 1,
+            &Self::Left(x) | &Self::Non(x) => x as u32 * 3 + 2,
+            &Self::Right(x) => x as u32 * 3 + 1,
         }
```
zesterer, line 503: This doesn't seem to match up with the table above it.

lancylot2004: Apologies I should've changed the table - the values in the table don't work either!

zesterer: Does this mean the values in the table are correct now? Apologies about not having gotten back to you in a while!

PR #729, opened 2025-02-07
chore: polish `ConfigParser`

yjhmelody:
No need to add `Self: Sized`.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -33,27 +33,6 @@ macro_rules! go_extra {
     };
 }
 
-macro_rules! go_cfg_extra {
-    ( $O :ty ) => {
-        #[inline(always)]
-        fn go_emit_cfg(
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -143,7 +122,6 @@ use self::{
     input::{
         BorrowInput, Emitter, ExactSizeInput, InputRef, MapExtra, SliceInput, StrInput, ValueInput,
     },
-    inspector::Inspector,
```
zesterer, line 146: Looks like this is causing CI failures.

zesterer (changes requested):
Thanks for the changes. Looks like CI needs fixing.

zesterer:
And a `cargo fmt`, it looks like

zesterer:
Thanks so much!

PR #733, opened 2025-02-21
Add a combinator `then_or_not` to optionally run a parser after another

### LINKED ISSUE #630: Combine `then` and `or_not` preventing extraneous backtracking
I have parsers `a` and `b` and I want the following rules to work:
  - if `a` succeeds, we try to parse `b` and the combined parser returns an error from `b` or `Some(result_tuple)`
  - if `a` fails, the combined parser returns `None`

If you just do `a.then(b).or_not()`, which will have the desired return type, `or_not` swallows both errors, but I need it to only swallow the error from `a`.

An example where this is useful (ignore_then instead of then but the idea's the same):

```rust
// d(x) = just(x).padded()

let param = ident().then(d("=").ignore_then(simple_val).or_not());

let func = keyword("fn").padded().then(ident())
    .then(param
        .separated_by(d(","))
        .collect()
        .delimited_by(d("("), d(")")))
```

Now trying to parse `fn test(a, b = bad_val)` you'll get an unhelpful error 'found = expected )', and it does make sense that it does that, but there should be a way to bubble up the error produced by the simple_val parser when we're gone past the `=`.

I haven't found a good way of doing this short of trying to write my own primitive - is there such a way?.

If there was a flat_map of sorts (I vaguely remember something like that from scala combinators) where you could do
`d("=").or_not().flat_map(|value| if value.is_some() { simple_val } else { noop } )`
it'd be an okay-ish hack I guess.

Like the doc of the _ctx methods says `Parse one thing and then another thing, creating the second parser from the result of the first.`, but it doesn't.. exactly do that?. There's this whole `configure` thing instead.

---

Okay, since writing the above and leaving it in a tab I've found out `custom` exists and wrote this
```rust
fn if_prefix<'a, I, E, X, O>(
    prefix: impl Parser<'a, I, X, E> + Clone + 'a,
    parser: impl Parser<'a, I, O, E> + Clone + 'a,
) -> impl Parser<'a, I, Option<O>, E>
where
    I: Input<'a>,
    E: ParserExtra<'a, I>,
{
    custom(move |input| {
        let marker = input.save();
        Ok(match input.parse(prefix.clone()) {
            Ok(_) => Some(input.parse(parser.clone())?),
            Err(_) => {
                input.rewind(marker);
                None
            }
        })
    })
}
```
which seems to work great.

Still posting this issue because I think this (or a bit more generalized version of this) is useful enough to be added to the lib.

Actually even in nanorust (if it had return types), `if_prefix(token("->"), return_type)` would've been useful.

---

Also also, the same issue exists with separated_by:
parsing `(a,a,a,bad)` gets us an error 'found , expected )', while not backtracking after a valid separator was parsed and pointing to the `bad` token would've been better (perhaps having a config similar to how there's allow_trailing)

91khr:
Closes #630, which first mentioned something like this.

This can be used in syntaxes like Rust's let statement, which has an optional type notation:

```rust
let a = 0;
let b: i32 = 0;
```

`.then(...).or_not()` doesn't work well in such cases, because when there's a parse error in the type, `or_not` would just suppress it and return `None`, leaving the following parsers confusing, or even worse, mistakenly starting parsing from the wrong position. In this particular example, when the type has an error, the parser may complain something like `expected '=', got ':'`, rather than pointing out the error in the type.

With `then_or_not`, Rust's let statement can roughly be parsed with:

```rust
let let_stmt = group((
        just("let").padded(),
        pattern,
        just(':').padded().then_or_not(rust_type),
        just('=').padded().then_or_not(expr),
    ))
    .map(|(_, pat, ty, val)| Stmt::Let(pat, ty.map(|t| t.1), val.map(|v| v.1)));
```

And this would correctly report the errors in `rust_type` and `expr`.

### DIFF
--- a/src/combinator.rs
+++ b/src/combinator.rs
@@ -2753,6 +2753,81 @@ where
     go_extra!(U);
 }
 
+/// See [`Parser::then_or_not`]
+pub struct ThenOrNot<A, B> {
+    pub(crate) parser_a: A,
+    pub(crate) parser_b: B,
+}
+
+impl<A: Copy, B: Copy> Copy for ThenOrNot<A, B> {}
+impl<A: Clone, B: Clone> Clone for ThenOrNot<A, B> {
+    fn clone(&self) -> Self {
+        ThenOrNot {
+            parser_a: self.parser_a.clone(),
+            parser_b: self.parser_b.clone(),
+        }
+    }
+}
+
+impl<'src, I, T, U, E, A, B> Parser<'src, I, Option<(T, U)>, E> for ThenOrNot<A, B>
+where
+    I: Input<'src>,
+    E: ParserExtra<'src, I>,
+    A: Parser<'src, I, T, E>,
+    B: Parser<'src, I, U, E>,
+{
+    fn go<M: Mode>(&self, inp: &mut InputRef<'src, '_, I, E>) -> PResult<M, Option<(T, U)>>
+    where
+        Self: Sized,
+    {
+        let before = inp.save();
+        let Ok(pre) = self.parser_a.go::<Emit>(inp) else {
+            inp.rewind(before);
+            return Ok(M::bind(|| None));
+        };
+        let post = self.parser_b.go::<M>(inp)?;
+        Ok(M::combine(M::bind(move || pre), post, |a, b| Some((a, b))))
+    }
+
+    go_extra!(Option<(T, U)>);
+}
+
+impl<'src, I, T, U, E, A, B> IterParser<'src, I, (T, U), E> for ThenOrNot<A, B>
+where
+    I: Input<'src>,
+    E: ParserExtra<'src, I>,
+    A: Parser<'src, I, T, E>,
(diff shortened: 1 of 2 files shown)


zesterer:
> .then(...).or_not() doesn't work well in such cases, because when there's a parse error in the type, or_not would just suppress it and return None, leaving the following parsers confusing, or even worse, mistakenly starting parsing from the wrong position

I'm not sure I follow this logic. Optional type annotations can be expressed perfectly well with

```rs
let pat = // Pattern
let ty = // Type annotation

text::keyword("let").padded()
    .ignore_then(pat)
    .then(just(":").padded().ignore_then(ty).or_not())
    .then_ignore(just("=").padded())
    .then(...)
```

Chumsky doesn't not get confused by this, and the fact that the error is suppressed when parsing fails is not a problem: chumsky will still remember that the error *could* have been hit and will make use of that information when generating the final syntax error.

zesterer:
Is this still an API hole for you, or is my suggested solution sufficient?

Issue #747, opened 2025-03-27
How to do the following things in 0.10.0


muja:
Hi, I've seen [this issue](https://github.com/zesterer/chumsky/issues/745), would you prefer it if we used that as a "megathread" or create new issues?

Anyway I am struggling with the following:

- `take_until` was removed and I couldn't find any adequate replacement. There is `none_of.repeated` but I don't understand how to make it work for multiple chars. For instance, how would you write this in 0.10.0? `just("/*").then_ignore(take_until("*/"))`
- `error::Simple` seems to not be able to be constructed? I've tried my hand at the `expected_found` function but I gave up after a million type errors. How would I migrate this function?
  ```rust
  fn ident() -> impl Parser<char, (Token, Span), Error = Simple<char>> {
      text::ident().try_map_with(|ident: &str, e| {
          if ident == "RESERVED1" || ident == "RESERVED2" {
              Err(Simple::custom(span, "RESERVED* keywords not allowed here"))
          } else {
              Ok((Token::Ident(ident.to_owned()), e.span()))
          }
      })
  }
  ```

muja:
And for the first one: I don't think `delimited_by` will cut it.

I'm doing string interpolation, so the exact 0.9.x logic is:

```rust
fn interpolated_string() -> impl Parser<char, Vec<(Token, Span)>, Error = Simple<char>> {
    just('"').ignore_then(recursive(|value| {
        take_until(just("\"").or(just("@(")).rewind())
            .map(|(chars, _)| chars)
            .collect()
            .map_with_span(|tok, span| (Token::RawString(tok), span))
            .chain(
                just("\"").map(|_| Vec::<(Token, Span)>::new()).or(just("@(")
                    .map_with_span(|_, span| (Token::InterpolationStart, span))
                    .chain(expression())
                    .chain(just(')').map_with_span(|_, span| (Token::InterpolationEnd, span)))
                    .chain(value)),
            )
    }))
}
```

zesterer:
`take_until` was removed because its use was ambiguous. It's been replaced by several other combinators that can be used in combination (hence the name!) to produce the same or similar effects. They are:

- `any()`, which already existed
- `a.and_is(b)`, which parses an `a` if `b` would *also* be matched in the same place
- `a.not()`, which parses nothing successfully (like `empty`), but *only* if `a` is not matched

[...]

PR #760, opened 2025-04-13
fix(examples): configure ariadne to use byte indices

muja:
Relates-to #747


zesterer:
Thanks! I'd forgotten these hadn't been changed too.

PR #761, opened 2025-04-13
added nom8 benches

asibahi:
Hello.

I saw `nom` was still at 7, so I added `nom 8` as a separate bench for `cbor` and `json`. Didn't change the existing logic otherwise.


zesterer:
Thanks! Looks like it needs a `cargo fmt` :)

Issue #774, opened 2025-04-23
Lack of documentation on integrating and composing chumsky errors


pehdfms:
First of all I want to give some appreciation for the work in the recent 0.10.0 release! It happened to align with a new project I've started working on and wow it's been a treat to get to use it in practice so soon. Unfortunately this alignment does mean that I got caught in the "not quite there yet" phase of documentation, especially when related to errors.

Let's look at a simple lexer variable:
```rust
let string = none_of('"') // TODO: Handle escapes.
    .repeated()
    .to_slice()
    .map(ToString::to_string)
    .delimited_by(
        just('"'),
        just('"').map_err(|err: Rich<char>| -> Rich<char> {
            Rich::custom(err.span().clone(), "Unterminated string literal")
        }),
    )
    .boxed();
```

This works fine, I suppose. But err.span().clone() feels just a bit dirty (I know, I know, it's a cheap type, let's just explore a little), so I'm thinking it'd be cool to just "edit" the error label that already exists, so let's do that:

```rust
// Zoom in to map_err
|err: Rich<char>| -> Rich<char> {
    err.label_with("Unterminated string literal") // labelled() returns in place, so this feels appropriate, right?
}
```

Ah, but that doesn't quite work. err.label_with returns an empty tuple (as rust helpfully points out!), so it's meant to be modifying the err in place. Let's try just modifying it and returning immediately then.

```rust
|mut err: Rich<char>| -> Rich<char> {
    err.label_with("Unterminated string literal"); // Type annotation needed
    err
}
```

And that explodes. Let's try to annotate it naively. We're returning a Rich error which parses over char, as explained in the function definition, right? So let's try to just type it as what Rust suggests:

```rust
|mut err: Rich<char>| -> Rich<char> {
    <Rich<char> as LabelError<&str, &str>>::label_with(&mut err, "Unterminated string literal")
    err.label_with("Unterminated string literal."); // Type annotation needed
    err
}
```

There we go. A little bit more of a mouthful than I expected, but I suppose that's fine. I think it'd be nice to add expected / found info since we know we want a " to finish this string, so let's try the Rich::expected_found() constructor.

```rust
|mut err: Rich<char>| -> Rich<char> {
    let new_err = <Rich<char> as LabelError<&str, char>>::expected_found(
        vec!['"'],
        err.found().map(|c| Maybe::Val(*c)),
        err.span().clone(),
    );
    new_err
}
```

Now let's check what that error looks like on our little suite of tests:
```
[...]

PR #775, opened 2025-04-23
Update Errors and Recovery Documentation

pehdfms:
As mentioned in #774, I'm working on updating errors and recovery docs with the info I find. This will be a draft for now, as I'll be needing some pointers on maintaining the writing style of the rest of the documentation as well as tips on extra detail to include.

### DIFF
--- a/guide/error_and_recovery.md
+++ b/guide/error_and_recovery.md
@@ -81,8 +81,8 @@ pub fn lexer<'src>() -> impl Parser<'src, &'src str, Expr, extra::Err<Rich<'src,
             .map(Expr::List)
             // If parsing a list expression fails, guess at the error being accidental text input.
             .recover_with(via_parser(text_recovery.map(Expr::Error)))
-            // Fallback to less conservative recovery otherwise.
             .or(int)
+            // Fallback to less conservative recovery otherwise.
             .recover_with(via_parser(naive_recovery.map(Expr::Error)))
             .padded()
     })
--- a/guide/error_and_recovery.md
+++ b/guide/error_and_recovery.md
@@ -43,16 +43,74 @@ In the next section we'll see how to define custom Recovery Strategies for our p
 A Recovery Strategy is nothing more than a parser expression to be called in exceptional conditions, Consider the following code:
 
 ```
-# use chumsky::prelude::*;
-# fn parser<'src>() -> impl Parser<'src, &'src str, ()> { end() }
-#[test]
-fn test_parser() {
-    // Our parser expects empty strings, so this should parse successfully
-    assert_eq!(parser().parse("").into_result(), Ok(()));
-
-    // Anything other than an empty string should produce an error
-    assert!(parser().parse("123").has_errors());
+#[derive(Clone, Debug, PartialEq, Eq)]
+enum ErrorKind {
+    UnexpectedText(String),
+    Unknown,
 }
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+enum Expr {
+    Error(ErrorKind),
+    Int(i64),
+    List(Vec<Expr>),
+}
+
+pub fn lexer<'src>() -> impl Parser<'src, &'src str, Expr, extra::Err<Rich<'src, char>>> {
+    // Atoms.
+    let int = text::int(10).from_str().map(|r| match r {
+        Ok(int) => Expr::Int(int),
+        Err(_) => Expr::Error(ErrorKind::Unknown),
+    });
+
+    // Errors.
+    let text_recovery = any()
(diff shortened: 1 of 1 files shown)


guide/error_and_recovery.md
```diff
@@ -1,3 +1,61 @@
...
+
+## What is an [`Error`]?
+
+While Chumsky is responsible for the plumbing and detection of errors based on Parser Combinator outputs, you have full control over the creation, display, contained data and interactions for errors you define. Chumsky achieves this by having parser combinators automatically generate errors based on their parser function's type signature while providing the [`Parser::map_err`] utility to convert the original error type into your own.
```
zesterer, line 7: Hmm, I think this paragraph feels a bit fluffy and doesn't really say much. May I recommend:

> Chumsky manages the tracking and generation of errors for you. However, you may customise what information an error keeps track of and how it tracks it by changing the error type (see [`extra::Err`]). Chumsky provides a range of built-in error types to handle a range of use-cases, but you can also implement the [`Error`] trait for your own error type if you want even more fine-grained control.

In my view, talking about `map_err` is probably a point for another heading - it, along with combinators like `try_map`, `validate`, etc. are more in the realm of error manipulation rather than actually defining properties of errors.

guide/error_and_recovery.md
```diff
@@ -1,3 +1,61 @@
...
+
+- [`Cheap`]: contains only [`Span`] information about the associated error, letting you know where it happened, but not giving you much more information besides. As the name implies, it is very cheap, and thus usable for high performance environments where basic usability is still nice to have.
+
+- [`Simple`]: contains the [`Span`], but also the found token at the location of the error. This error is a good middle ground between the raw speed of [`Cheap`] and the detailed user-centric focus of [`Rich`], while still being able to satisfy medium performance requirements.
+
+- [`Rich`]: the recommended type for dealing with user input at the expense of raw speed. Rich is still *fast* for applications like compilers and interpreters, and a great choice if you want to give users a pleasant editing experience first and foremost. Rich error types contain [`Span`] information as well as found and expected values, a [`Label`] to describe the associated parser construction, the ability to merge errors in the same span, as well as a potential [`Context`] with truly arbitrary data.
+
+## The [`Span`] Trait
+TODO
```
zesterer, line 24: For this paragraph I might recommend:

> Spans are an important feature of any competent parser. They provide a location within a source file, denoted with start and end offsets, that corresponds to the location of some element within the source code (often a single character or token, but also an entire multi-token high-level structure like an expression, statement, or function).

> Relevant to this section, spans are also used as an input to parser errors, allowing the final message shown to the user to reference the location at which an error occurred, along with other contextual information (such as which structure the error occurred within).

zesterer (commented):
Thanks so much for working on this! I like the writing style. I've added a few suggestions. Funny thing: I started writing docs for just this a few days ago (I've not committed them yet). I think I prefer the way yours is written, but it's curious that there are so many parallels!

pehdfms:
> I've not had time to review everything (I'm writing this on a train right now), but hopefully I'll get the time to finish off tonight or tomorrow.
Don't worry about it, I'm not in any rush here - I'm recovering from surgery and have a lot of free time 

Don't worry about it! I'm currently recovering from surgery so I might not be extremely responsive myself. I've noted your suggestions and will try to work on it later today, thank you for your time :)

pehdfms:
Wrote a little more today, I think the example I provided for a recovery strategy might be a little too extensive for the section its in (it does show a few different concepts at once), but I'm curious about your thoughts.

I sprinkled a couple TODO comments asking questions in the commit, but they (and a couple additional ones) are entirely summarized in the list below:

- I have not been able to discover what the "fallback" parameter of skip_until is. What is it used for? Is it equivalent to chaining a new recover_with?
- Need some pointers on writing for skip_then_retry_until. I believe I had seen some of it defined elsewhere in the docs before, but I haven't been able to find it again today, and trying to write by memory just left me blank.
- Unsure about how much we should write about ariadne in this page. I think at least a link to the docs (and maybe one example on how to handle automated diagnostics) would be nice. Having a more dedicated guide might be even nicer, but I'm unsure if that's out of the scope of Chumsky's docs.
- I'm lacking a bit of intuition on recover_with. When it's called on a parser, does it resume parsing where the previous one stopped, or does it entirely restart? The distinction can feel a little difficult to grasp in practice when working with recursive descent parsers since each individual parser span is chopped so thin.
- What's next? I suppose we shouldn't yet include a section on the RecoveryStrategy trait since it's sealed. Am I missing any additional traits that may be relevant? Some other function that I haven't noticed yet? Feel free to comment.

PR #777, opened 2025-04-26
feat(nightly): add conversions and Span impl for new range api

muja:


zesterer:
Nice! Glad to see upstream progress on the original issues with Range.

muja:
yeah they are a lot easier to use. Do you have any concerns with merging this? I've been using my fork successfully

zesterer:
No concerns, seems like GitHub CI isn't running for it though. I'll check back tomorrow to see if it's a temporary issue.

PR #800, opened 2025-05-22
Add non-associative infix operators

NomAnor:
I tried to mess with the code and got a working solution. I'm new to rust, so this might not be the best code.

I had to disallow a precedence of zero, which might be a breaking change. Zero is used as a start marker
and all implementations I've seen use precedences > 0 when defining the operators.

I also switched the check for associtivity with the operator parsing, which might change the performance?

Now the test `pratt::tests::with_pre_and_postfix_ops` fails because it contains an ambiguity between the `§` and `+` operator, both have a binding power of 4.
The comment in that test case indicates that such ambiguity is not a good thing, so maybe this new behaviour is okay? Then the check could just be changed. The old behaviour bound infix operators tighter than prefix operators of the same precedence.

This is a new attempt, the other ones are #600 and #728.

### DIFF
--- a/src/pratt.rs
+++ b/src/pratt.rs
@@ -444,6 +444,8 @@ pub enum Associativity {
     Left(u16),
     /// Specifies that the operator should be right-associative, with the given precedence (see [`right`]).
     Right(u16),
+    /// Specifies that the operator is non-associative, with the given precedence (see [`none`]).
+    None(u16),
 }
 
 /// Specifies a left [`Associativity`] with the given precedence.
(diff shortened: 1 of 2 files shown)


src/pratt.rs
```diff
@@ -428,52 +438,67 @@ where
...
 
-/// Specifies a right [`Associativity`] with the given binding power.
+/// Specifies a right [`Associativity`] with the given precedence.
 ///
 /// Right-associative operators are evaluated from the right-most terms, moving leftward. For example, the expression
 /// `a ^ b ^ c ^ d` will be evaluated as `a ^ (b ^ (c ^ d))` because exponents are conventionally right-associative.
-pub fn right(binding_power: u16) -> Associativity {
-    Associativity::Right(binding_power)
+pub fn right(precedence: u16) -> Associativity {
+    assert!(precedence > 0);
```
zesterer, line 475: Might be worth adding `#[track_caller]` to this function (and others), which should result in the assert location pointing to the caller location rather than inside the function.

zesterer (changes requested):
This is very cool, thanks so much for your work! I've just a small suggestion above.

Additionally, I'm wondering whether we could avoid breaking binding powers of 0 by casting the `u16` up to a `i32` and using `i32::MIN` as the starting marker?

zesterer:
Hmm, looks like there are CI failures too.

NomAnor:
> I have to think how to fix that.

Got it, just need to check if the operator is actually non-associative.

What do you thing about the error handling/message?

NomAnor:
I haven't wrapped my head around the error handling. If I need to use it in my code, I might come back to it.
For now I think the base functionality is there.

zesterer:
Thanks! I may make some minor modifications to this post-merge, but I really appreciate all of the work you've put into this!

PR #806, opened 2025-05-25
feat: add `Parser::try_foldl` combinator

### LINKED ISSUE #805: failable shortcircuiting fold
Is there a way to have a short circuiting fold operation? Similar to [`std::Iterator::try_fold`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.try_fold)? I can do `foldl` on an `Option` but that will not stop once the reducing operation encountered an error.

Swiftb0y:
This commit implements a `TryFoldL` parser along with the `Parser::try_foldl` combinator. It allows left folding the output like `Parser::foldl` put is able to fail during the reduction.

I have left in two `TODO`s where I would like to receive some feedback on.

Closes #805

### DIFF
--- a/src/combinator.rs
+++ b/src/combinator.rs
@@ -558,29 +558,28 @@ where
     A: Parser<'src, I, OA, E>,
     B: IterParser<'src, I, OB, E>,
     E: ParserExtra<'src, I>,
-    F: Fn(OA, OB, I::Span) -> Result<OA, E::Error>,
+    F: Fn(OA, OB, &mut MapExtra<'src, '_, I, E>) -> Result<OA, E::Error>,
 {
     #[inline(always)]
     fn go<M: Mode>(&self, inp: &mut InputRef<'src, '_, I, E>) -> PResult<M, OA>
     where
         Self: Sized,
     {
+        let before_all = inp.cursor();
         let mut out = self.parser_a.go::<Emit>(inp)?;
         let mut iter_state = self.parser_b.make_iter::<Emit>(inp)?;
         loop {
-            // TODO what span is appropriate to pass here?
-            // only from this iteration or since the beginning?
             let before = inp.cursor();
             match self.parser_b.next::<Emit>(inp, &mut iter_state) {
                 Ok(Some(b_out)) => {
-                    let span = inp.span_since(&before);
-                    match (self.folder)(out, b_out, span) {
+                    match (self.folder)(out, b_out, &mut MapExtra::new(&before_all, inp)) {
                         Ok(b_f_out) => {
                             out = b_f_out;
                         }
-                        // TODO properly report error
-                        // how does the `Errors` struct work (alt, secondary)?
-                        Err(_) => break Err(())
+                        Err(err) => {
+                            inp.add_alt_err(&before.inner, err);
+                            break Err(());
+                        },
                     }
                 }
                 Ok(None) => break Ok(M::bind(|| out)),
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -748,7 +748,7 @@ pub trait Parser<'src, I: Input<'src>, O, E: ParserExtra<'src, I> = extra::Defau
     ///
     /// let sum = int
     ///     .clone()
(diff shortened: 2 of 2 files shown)


src/combinator.rs
```diff
@@ -526,6 +526,80 @@ where
...
+    }
+}
+
+impl<'src, I, F, A, B, OA, OB, E> Parser<'src, I, OA, E> for TryFoldl<F, A, B, OB, E>
+where
+    I: Input<'src>,
+    A: Parser<'src, I, OA, E>,
+    B: IterParser<'src, I, OB, E>,
+    E: ParserExtra<'src, I>,
+    F: Fn(OA, OB, I::Span) -> Result<OA, E::Error>,
```
zesterer, line 561: `I::Span` shouldn't be passed in here. If you need a span, it's better to also create a `TryFoldlWith` combinator. Like the other `XWith` combinators, it generalises to accessing spans, context, and state.

Additionally, there's already a clue in `FoldlWith` regarding the span: the span starts from before everything (`before_all`).

src/combinator.rs
```diff
@@ -526,6 +526,80 @@ where
...
+            match self.parser_b.next::<Emit>(inp, &mut iter_state) {
+                Ok(Some(b_out)) => {
+                    let span = inp.span_since(&before);
+                    match (self.folder)(out, b_out, span) {
+                        Ok(b_f_out) => {
+                            out = b_f_out;
+                        }
+                        // TODO properly report error
+                        // how does the `Errors` struct work (alt, secondary)?
+                        Err(_) => break Err(())
```
zesterer, line 583: The rule is that if a parser returns an `Err(())`, `inp` should have been made aware of the details of at least one error (right down at the base of the crate, in `Parser::parse`, there's effectively an `.unwrap()` on the error information so not providing anything would cause a problem).

My sense is that we should probably do the same thing here that we do for `repeated` - `TryFoldl` should always return `Ok(_)` (provided the initial element was parsed, of course) and should instead just record the error. That means that we should do something like this:

```rs
Err(err) => {
    inp.add_alt_err(&before.inner, err);
    break Err(());
},
```

zesterer (changes requested):
Thanks for the PR! I think the broad strokes are all reasonable. I've added a comment for each `TODO`. You might want to consider adding a `TryFoldlWith` combinator (for completeness, and for your span use-case).

src/combinator.rs
Swiftb0y: I partly based this of `try_map` and `foldl`. In `try_map` we do get a span, but of course I can split that to a `TryFoldlWith` and do it similar to `foldl_with` if you prefer.

src/combinator.rs
Swiftb0y: right. My primary issue is that I don't really understand what the `InputRef::add_alt_err` function or the `input::Errors` struct do. Admittedly I haven't bothered to look at `InputRef` very deeply yet.

src/combinator.rs
zesterer: Hmm, I do understand this reasoning on a second pass. It's not possible to generate an error without having access to the span anyway. Perhaps `try_foldl` should just include the extra parameter (i.e: like `foldl_with`) by default?

Swiftb0y: I'm not sure. I like the consistency with `try_map`, but I'm not sure if its worth to implement a second entire `try_foldl_with`. I've taken the easy route for now.

src/combinator.rs
Swiftb0y: Thanks for the explanation, that makes very much sense.

zesterer (approved):
Thanks so much! I think this is quite a valuable change. +1 for adding the example too, that's great.

Swiftb0y:
Yikes, this contained a `fixup!` I would've preferred to squash before merging. Anyways, thanks for the your explanations, fast review and inclusion.

zesterer:
Eh, I think that history is often useful. I'm not generally a fan of squashing. If you really want I can retroactively squash the history though?

Swiftb0y:
in that case I would've given it a better description than just `fixup`. I wouldn't squash now since that would force to force push to a public branch... Just leave it. sorry for the complications.

PR #808, opened 2025-05-29
Update indent.rs in order to have a proper Pratt parser

Germ210:
This rewrites indent.rs to use a real Pratt parser. Previously, the file had a TODO noting that it should eventually use expression parsing, but in the meantime it didn’t actually implement any precedence rules—it just matched the text "expr" directly. This change introduces proper precedence handling using a Pratt parser, making the indentation-sensitive parser more accurate and extensible, while keeping the overall behavior consistent.

### DIFF
--- a/examples/indent.rs
+++ b/examples/indent.rs
@@ -1,25 +1,61 @@
 use chumsky::prelude::*;
+use chumsky::pratt::*;
+
+#[derive(Clone, Debug)]
+pub enum Expr {
+    Add(Box<Self>, Box<Self>),
+    Sub(Box<Self>, Box<Self>),
+    Mul(Box<Self>, Box<Self>),
+    Div(Box<Self>, Box<Self>),
+    Pow(Box<Self>, Box<Self>),
+    Neg(Box<Self>),
+    Factorial(Box<Self>),
+    Deref(Box<Self>),
+    Literal(i32),
+}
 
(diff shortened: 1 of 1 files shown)


zesterer:
Hi, looks like you need to enable the `pratt` feature for the `indent` example in `Cargo.toml`.

examples/indent.rs
```diff
@@ -32,17 +68,14 @@ fn parser<'a>() -> impl Parser<'a, &'a str, Vec<Stmt>> {
...
-    expr
-    loop:
-        expr
-        expr
-    expr
-expr
+  *1 + -2 * 3!
+  loop:
+        10
+
```
zesterer, line 76: It's important to also show that trailing expressions work here too (i.e: expressions that appear after a nested block). I'd suggest something like the following example:

```
loop:
    1 + 2 * 3
    loop:
        10 + 3
    42 * 3
3 + 2 * 9
```

Germ210: Alright,  I can add that later, but in the original parser (and presumably this one), when given the input: 
`expr
expr
loop:
expr
`

It will output the tree:
`
Some(
    [
        Expr,
        Expr,
        Loop(
            [
                Expr,
            ],
        ),
    ],
)
`

Is this the intended behavior? I would assume that in order to be nested, it would need to have the corresponding indentation? Or is this intentional?

Issue #809, opened 2025-05-29
make `Rich` compare spans instead of all members for `Eq`


alex-s168:
from the current code:
```rs
#[derive(Clone, PartialEq, Eq, Hash)]
pub struct Rich<'a, T, S = SimpleSpan<usize>> {
    span: S,
    reason: Box<RichReason<'a, T>>,
    context: Vec<(RichPattern<'a, T>, S)>,
}
```

this is annoying because `T` might not always implement `Eq` and `PartialEq`

for example `proc_macro2::TokenTree` doesn't

alex-s168:
> Is there something specific you need `PartialEq` for?

No, but `LabelError` needs `PartialEq` on tokens

zesterer:
I am happy to accept a PR for this change!

PR #811, opened 2025-06-01
Make Rich's PartialEq only compare the span

alex-s168:
#809 

No idea what I'm doing TBH, but all the tests pass


alex-s168:
turns out I'm stupid and LabelError needs PartialEq on tokens...

Issue #812, opened 2025-06-04
`chumsky::text::Padded` and `chumsky::text::whitespace` make parsers vulnerable to CVE-2021-42574


TurtleArmyMc:
CVE-2021-42574 describes an issue where some Unicode characters can be used to create source code that looks ok to the reader, but is parsed in an expected way. Specifically, it's when someone uses whitespace characters that change text direction from left-to-right/right-to-left while being invisible to reorder code visually. Chumsky's `.padded()` and `text::whitespace()` parsers only check if a character is whitespace, which does exactly what they promise, but this leaves parsers vulnerable to this CVE. There might be other places in chumsky that also has this issue that I missed.

Some relevant links about the CVE.
https://trojansource.codes/
https://nvd.nist.gov/vuln/detail/CVE-2021-42574
https://blog.rust-lang.org/2021/11/01/cve-2021-42574/

Here's a (somewhat) minimal example of what this could look like for parsing a subset of C code. Here it looks like `src` checks for admin permissions before calling `adminOnlyCode`, but chumsky will parse this as JUST the line `adminOnlyCode();` padded by a bunch of block comments and whitespace.
```rust
use chumsky::prelude::*;

// This code contains invisible evil characters! Rust 1.56.1 onwards protects us from them by default.
#[allow(text_direction_codepoint_in_literal)]
let src = r#"
    /* The entire line below this one is actually commented */
    /*‮ } ⁦if (isAdmin)⁩ ⁦ begin admins only */
        /* Chumsky will parse this string as just this one function call padded by comments/whitespace */
        adminOnlyCode();
    /* The line below this one is also just a comment */
    /* end admins only ‮ { ⁦*/
"#;

#[derive(Debug, PartialEq, Eq)]
enum Token<'src> {
    If {
        condition: &'src str,
        then: Vec<Token<'src>>,
    },
    Call(&'src str),
}
let comments = any::<_, extra::Err<Simple<char>>>()
    .and_is(just("*/").not())
    .repeated()
    .delimited_by(just("/*"), just("*/"))
    .padded() // This innocent line is what makes the parser vulnerable
    .repeated();
let mut tokens = Recursive::declare();
let if_ = just("if")
    .padded_by(comments)
    .ignore_then(
        text::ident()
            .padded_by(comments)
            .delimited_by(just('(').padded_by(comments), just(')').padded_by(comments)),
    )
    .padded_by(comments)
    .then(
        tokens
            .clone()
[...]

PR #813, opened 2025-06-06
fix: #812 Reject whitespace characters vulnerable to CVE-2021-42574

TurtleArmyMc:
I haven't updated any of the documentation in `chumsky::text::whitespace` or any of the other related user facing API documenting that some whitespace characters are rejected, but the fix is in place. Should I update the documentation too?

Unit tests would be a good idea too, but I'm not sure where the best place to put them would be.

The tutorials also have the stdlib `char::is_whitespace` in a couple of places, but I feel like updating those calls to `chumsky::text::unicode::is_whitespace` would just be extra noise getting in the way of people understanding what the tutorials are trying to teach, even if it does mean that people might be more likely to have vulnerable parsers because of following the tutorial.

### DIFF
--- a/src/text.rs
+++ b/src/text.rs
@@ -49,7 +49,7 @@ impl Char for &Grapheme {
 
     fn is_whitespace(&self) -> bool {
         let mut iter = self.as_str().chars();
-        iter.all(char::is_whitespace)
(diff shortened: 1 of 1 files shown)


zesterer:
Thanks!

PR #816, opened 2025-06-26
fix: typos

hackaugusto:
Fix some small typos.

### DIFF
--- a/guide/getting_started.md
+++ b/guide/getting_started.md
@@ -131,7 +131,7 @@ fn test_parser() {
 ```
 
 Hopefully, this code is fairly self-explanatory. We call `parser()` (the function we wrote in the previous section) to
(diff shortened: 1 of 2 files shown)


guide/getting_started.md
```diff
@@ -130,7 +130,7 @@ fn test_parser() {
 }
 ```
 
-Hopefully, this code is fairly self-explanatory. We call `parse()` (the function we wrote in the previous section) to
+Hopefully, this code is fairly self-explanatory. We call `parser()` (the function we wrote in the previous section) to
 create an instance of our parsers, and then we call [`Parser::parse`] on it with the desired input to actually do some
```
zesterer, line 134: I've just noticed that 'parsers' in the following line should be 'parser'.

hackaugusto: updated :)

zesterer:
Thanks!

PR #817, opened 2025-06-26
fix: generic changed from Container to OrderedSeq

hackaugusto:
The generic name [used to be `C: Container<I>`](https://github.com/zesterer/chumsky/commit/b38db3e1e5ef6e94b690b08fdb77064d98ff3823) but it changed to `T: OrderedSeq<I::Token> + Clone`


zesterer:
Thanks!

Issue #824, opened 2025-06-29
`not` leads to incorrect operation of `labelled`


Hedgehogo:
Test content:
```rs
use chumsky::extra::Err;
use chumsky::label::LabelError;
use chumsky::DefaultExpected;

fn parser<'src>() -> impl Parser<'src, &'src str, (), Err<Rich<'src, char>>> {
    just("b").not().labelled("label").ignored()
}

assert_eq!(
    parser().parse("b").into_output_errors(),
    (
        None,
        vec![{
            let mut err = LabelError::<&str, _>::expected_found(
                [DefaultExpected::SomethingElse],
                Some('b'.into()),
                SimpleSpan::new((), 0..1),
            );
            LabelError::<&str, _>::label_with(&mut err, "label");
            err
        }]
    )
);
```

Execution result:
```
assertion `left == right` failed
  left: (None, [found ''b'' at 0..1 expected something else])
 right: (None, [found ''b'' at 0..1 expected label])
```

Hedgehogo:
It seems that [this](https://github.com/zesterer/chumsky/blob/79b2ab35d44847441143b2a8e2312d088c495091/src/combinator.rs#L2292) line moves the cursor, which creates an error with the cursor not at the beginning of the input and labeled does not work.

PR #825, opened 2025-06-30
fix: `not` generates the wrong location for the error (#824)

Hedgehogo:


src/input.rs
```diff
@@ -1635,6 +1635,19 @@ impl<'src, 'parse, I: Input<'src>, E: ParserExtra<'src, I>> InputRef<'src, 'pars
...
+    where
+        I: ValueInput<'src>,
+    {
+        // SAFETY: cursor was generated by previous call to `Input::next`
+        let token = unsafe { I::next(self.cache, &mut self.cursor.clone()) };
+        if let Some(t) = &token {
+            self.state.on_token(t);
+        }
+        token
+    }
```
zesterer, line 1649: This doesn't look correct. `on_token` should only be called when a parser finally decides to consume a token, and so is not appropriate for a peeking operation. It seems like `InputRef::peek` should be sufficient for this, no?

Hedgehogo: Fixed it using `peek_maybe', not sure what the difference is, it seems even more correct?

Issue #827, opened 2025-07-01
Combination of the optional `try_map` and `separated_by' loses some of the expected value of the error


Hedgehogo:
Test content:
```rs
use crate::{DefaultExpected, LabelError};

fn tuple<'input>() -> impl Parser<'input, &'input str, (), extra::Err<Rich<'input, char>>> {
    group((
        any(), // It is necessary to avoid bug #799
        just("b").or_not(),
        any().try_map(|_, _| Ok(())).or_not(),
    ))
    .separated_by(just(","))
    .then_ignore(just(")"))
}

assert_eq!(
    tuple().parse("a").into_output_errors(),
    (
        None,
        vec![LabelError::<&str, _>::expected_found(
            vec![
                DefaultExpected::Token('b'.into()),
                DefaultExpected::Any,
                DefaultExpected::Token(','.into()),
                DefaultExpected::Token(')'.into()),
            ],
            None,
            SimpleSpan::new((), 1..1)
        )]
    )
);
```

Execution result:
```
assertion `left == right` failed
  left: (None, [found end of input at 1..1 expected any, '','', or '')''])
 right: (None, [found end of input at 1..1 expected ''b'', any, '','', or '')''])
```

PR #828, opened 2025-07-01
fix: `try_map` loses some of what is expected in error when used sequentially (#827)

Hedgehogo:
Fix for #827


zesterer (approved):
Makes sense to me, I'd missed that the `old_alt` ends up getting forgotten on failure. Thanks!

Hedgehogo:
> Let me know when you're ready to take this out of draught and merge.

I noticed that I haven't fixed `try_map_with`.

Issue #799, opened 2025-05-21
separated_by() crashes with an error when it shouldn't


Hedgehogo:
Minimum example:
```rs
        use chumsky::{extra, error::Simple};

        fn parser() -> impl Parser<'static, &'static str, (), extra::Err<Simple<'static, char>>> + Copy {
            just("a").or_not().separated_by(just("b"))
        }
        
        assert_eq!(parser().parse("bba").into_result(), Ok(()));
```

Gives out:
```
thread 'syntax::parse::whitespace::tests::test' panicked at C:\Users\Professional\.cargo\git\checkouts\chumsky-651fb76526ac39e3\11f5913\src\combinator.rs:1869:13:
found SeparatedBy combinator making no progress at src\syntax\parse\whitespace.rs:52:33
```

Hedgehogo:
I recommend looking at #836

PR #830, opened 2025-07-01
fix: `separated_by` crashes with an error when it shouldn't (#799)

Hedgehogo:


src/combinator.rs
```diff
@@ -2020,11 +2022,15 @@ where
...
-            );
+            if !first_loop {
+                debug_assert!(
+                    before != inp.cursor(),
+                    "found SeparatedBy combinator making no progress at {}",
+                    self.location,
+                );
+            } else {
+                first_loop = false;
+            }
```
zesterer, line 2033: Hmm. Maybe this logic should be moved to the `IterParser` implementation so it works for `.separated_by(...).collect()` too? What do you think?

Hedgehogo: > Hmm. Maybe this logic should be moved to the `IterParser` implementation so it works for `.separated_by(...).collect()` too? What do you think?

I've been thinking, I think we need the ability to call `next` by telling it whether to respond to a lack of progress.

PR #839, opened 2025-07-02
fix: `filter` creates an error with an incorrect `found`

Hedgehogo:


src/lib.rs
```diff
@@ -3817,6 +3817,17 @@ mod tests {
     fn filter() {
         use crate::{DefaultExpected, LabelError};
 
+        let parser = any::<_, extra::Err<Rich<_>>>().filter(|_| false);
```
zesterer, line 3820: Could you change this to `just("a")`? I think that would give us more assurance that it's the `Filter` that's generating this error and not the `any` (helps guard against future refactoring bugs).

src/combinator.rs
```diff
@@ -260,30 +261,28 @@ where
...
                     if let Some(new_alt) = new_alt {
                         inp.add_alt_err(&new_alt.pos, new_alt.err);
                     }
                     Ok(M::bind(|| out))
                 } else {
                     // If unsuccessful, reinsert the original alt but replace the new alt with the "something else" error (since it overrides it)
                     let expected = [DefaultExpected::SomethingElse];
-                    let err = E::Error::expected_found(expected, None, span);
-                    inp.errors.alt = old_alt;
+                    let err = E::Error::expected_found(expected, found, span);
```
zesterer, line 276: Hmm. I do understand the reasoning here: and I think it's the best we can do for now. But it isn't ideal long-term. Ideally we'd point at the *whole* of the matched input here instead of just the first token. Do you think you could add a TODO noting that this is something to improve on later?

Issue #843, opened 2025-07-05
`rewind` generates incorrect errors


Hedgehogo:
Test content:
```rs
use chumsky::{label::LabelError, DefaultExpected};

let parser = group((just("a"), any(), just("b").or_not()))
    .rewind()
    .then(just::<_, _, extra::Err<Rich<_>>>("ac"));

assert_eq!(
    parser.parse("ad").into_output_errors(),
    (
        None,
        vec![LabelError::<&str, _>::expected_found(
            [DefaultExpected::Token('c'.into())],
            Some('d'.into()),
            SimpleSpan::new((), 1..2)
        )]
    )
)
```

Execution result:
```
assertion `left == right` failed
  left: (None, [found end of input at 2..2 expected ''b''])
 right: (None, [found ''d'' at 1..2 expected ''c''])
```

Hedgehogo:
Fixed by #859

PR #844, opened 2025-07-05
fix: `rewind` generates incorrect errors (#843)

Hedgehogo:
This is a temporary fix that fixes the problem, but loses some of the error information (Which generally does not violate any chumsky principles, it does not guarantee the completeness of errors). 

The best solution would be to keep two `Cursors` inside `Location`, one referring to the current location, as usual, and the second to the last location outside of rewind. This also requires entering a separate marker where this location was. However, this is a major change that requires further discussion.


Hedgehogo:
Accepted together with #859.

PR #850, opened 2025-07-24
Add the Set combinator that applies a set of parser without order

peckpeck:
Add a `Set` combinator that takes a tuple of parsers and applies all of them in any order.

This permits parsing languages where arguments can be provided in any order.

`set((A, B))` parses A then B or B then A depending on which comes first.

Parsers that match without making progress are applied last to make sure they have no better alternative, which allows optional parsers like `A.or_not()`.


peckpeck:
Done, please tell me what you think

zesterer:
Looks great. One concern I have is that this parser always `Emit`s its output, which means missing out on one of the big optimisations that chumsky brings to the table, I think `go_or_emit` should be made generic over `M: Mode` to achieve this (hopefully other parsers explain how this works, but if not then I'm happy to help out).

Other than that, I'd be happy to merge this!

peckpeck:
Found it, and added some clippy fixes linked to MSRV

PR #855, opened 2025-08-08
feat: implement `ConfigParser` for `Filter` parser

Wybxc:
Allow the function of `Filter` to be configured based on context to enable more context-sensitive parsing.

Usage example:

```rust
// Filtering typedef names defined in the preceding context
identifier()
    .filter(|_| false) // Dummy filter to enable .configure
    .configure(|cfg, ctx| {
        let ctx = Rc::clone(ctx);
        cfg.filter(move |name| ctx.borrow().is_typedef_name(name))
    })
```

### DIFF
--- a/src/combinator.rs
+++ b/src/combinator.rs
@@ -259,22 +259,22 @@ where
 }
 
 /// Configuration for [`Filter`], used in [`ConfigParser::configure`]
-pub struct FilterCfg<F> {
-    pub(crate) filter: Option<F>,
+pub struct FilterCfg<O> {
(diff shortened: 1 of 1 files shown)


zesterer:
Hmm, I'm not sure how I feel about the boxing here. This code can run during parsing, and it's a potentially a rather high cost. Perhaps the type of the function could be lifted up from the `filter` method to the `FilterCfg` type?

zesterer:
Sure. Yes, that's definitely the better thing to use for your use-case. I'll close this PR for now.

Issue #862, opened 2025-08-12
`rewind` rolls back secondary errors


Hedgehogo:
Test content:
```rs
let parser = any::<_, extra::Default>()
    .validate(|out, _, emitter| {
        emitter.emit(EmptyErr::default());
        out
    })
    .rewind()
    .then_ignore(any());

assert_eq!(
    parser.parse("a").into_output_errors(),
    (Some('a'), vec![EmptyErr::default()])
);
```

Execution result:
```
assertion `left == right` failed
  left: (Some('a'), [])
 right: (Some('a'), [EmptyErr(())])
```

zesterer:
Seems like this was fixed by #863

PR #863, opened 2025-08-12
fix: combinators that should rewind only the input also rewind secondary errors (#862)

Hedgehogo:


zesterer:
Nice catch! I should have reviewed the original PR more carefully.

PR #866, opened 2025-08-15
feat: impl From<ExpectedNumber> for RichPattern

seftontycho:
Currently it is not possible to use the `number::Number` parsers with the `Rich` error type as `RichPattern` does not implement `From<ExpectedNumber>`.

This PR adds the missing impl in the `number` module so it is appropriatley feature gated.

Not previously possible:
```rust
fn parse_u8<'a>() -> impl Parser<'a, &'a str, u8, extra::Err<Rich<'a, char>>> {
    let u8_num: Number<STANDARD, &'a str, u8, ErrRich<'a>> = number();
    u8_num
}
```

### DIFF
--- a/src/number.rs
+++ b/src/number.rs
@@ -58,6 +58,12 @@ where
     go_extra!(O);
 }
 
+impl<'a, T> From<ExpectedNumber> for error::RichPattern<'a, T> {
+    fn from(_: ExpectedNumber) -> Self {
+        error::RichPattern::Label(std::borrow::Cow::Borrowed("number"))
+    }
+}
+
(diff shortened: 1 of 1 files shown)


seftontycho:
I'm not exactly sure what variant of `RichPattern` to use here, `Label` seemed the most fitting but then I'm not sure on the label name.

Open to suggestions better than "number".

zesterer:
Apologies, GitHub swallowed the notification for this.

I think the only change needed is to add `#[non_exhaustive]` to the `RichPattern` enum.

PR #875, opened 2025-09-09
relax Recursive::define() bounds

zer0guz:
``pub fn define<P: Parser<'src, I, O, E> + Clone + 'b + 'src>(&mut self, parser: P)``

this was the old signature and i just removed the ``+'src``

all tests passed and nothing else needed changing


zesterer:
Thanks!

PR #880, opened 2025-09-24
Add `MapExtra::emit(&mut self, E::Error)` method

zeichenreihe:
I still want to add tests.

### DIFF
--- a/src/input.rs
+++ b/src/input.rs
@@ -1892,6 +1892,8 @@ impl<'src, 'b, I: Input<'src>, E: ParserExtra<'src, I>> MapExtra<'src, 'b, I, E>
 
     /// Emits an non-fatal error.
     pub fn emit(&mut self, err: E::Error) {
-        self.emitted.secondary.push(Located::at(self.before.clone(), err));
+        self.emitted
+            .secondary
+            .push(Located::at(self.before.clone(), err));
     }
 }


src/input.rs
```diff
@@ -1840,20 +1851,23 @@ pub struct MapExtra<'src, 'b, I: Input<'src>, E: ParserExtra<'src, I>> {
     cache: &'b mut I::Cache,
     state: &'b mut E::State,
     ctx: &'b E::Context,
+    emitted: &'b mut Vec<E::Error>,
```
zesterer, line 1854: It would be preferable to have this field be a `&'b mut Errors<I::Cursor, E::Error>`, since this can be directly borrowed from `inp` in `MapExtra::new`, and allows the `MapExtra::emit` method to *directly* push the error - removing the cost of the vector.

zesterer (changes requested):

zesterer:
> I'd feel better with a test, given that `MapExtra` can appears at various places where emitting secondary errors is not done (as of before this PR). But I do not know how secondary errors are handeled. If they cannot change that a certain parsing path is already taken (so no backtracking on these secondary errors), I'm fine with not adding a test.

Secondary errors don't influence parsing, so I don't really think that a test is necessary. In effect, this is just an API change.

zeichenreihe:
Then it should be ready to merge now.

PR #894, opened 2025-10-27
Fix #486: Add another recursive parser

### LINKED ISSUE #486: Memory leak in `recursive` when using `define` function

Zij-IT:
This PR contains the suggested solution described in #494 for the problem described in #486, based on the solution provided by #664.

## Breaking Changes:
- `Recursive::define` and `Recursive::declare` removed from the public API
- `RecursiveN` and `recursive_n` added to the prelude

## Description

Introduced to the public API through this PR are the following items:
- `recursive_n` - a function which outputs a `RecursiveN` parser
- `RecursiveN` - a parser which contains a recursive parser, as well as all recursive dependencies as strong references
- `RecursiveArgs` - a trait which all arguments passed into `recursive_n` must fulfill. It is implemented up to a tuple of 8 parsers, as I hope for everyone that they never have to deal with 8 mutually recursive parsers.

These solve the memory leak by only allowing the user to make weak clones of the recursive parsers. When the parsers returned by `recursive_n` are dropped, all strong references are dropped. The user has no access to any of the strong references and therefor can't accidentally shoot themselves in the foot.

## Considerations

1. The documentation and examples of these elements could likely be improved. It's hard to come up with a compact example for mutually recursive parsers :sweat_smile:
2. Could `recursive` be replaced with `recursive_n`? In some cases it would unfortunately require the user to add a `: Recursive<_>` to their currently existing parsers. I understand _why_ Rust can't determine if the argument is a tuple given a name, or not, but it is surely unfortunate.

(Whoops. Looks like a got the wrong number for the commit message. dang it)

### DIFF
--- a/src/recursive.rs
+++ b/src/recursive.rs
@@ -260,9 +260,6 @@ pub trait RecursiveArgs<'src, 'b, I, O, E, P>: Sized {
     /// The return type of the parsers made by `Self`
     type Return;
 
-    /// A helper type for the macro :D Don't worry.
-    type Aux;
-
     /// Define the recursive parsers, returning a tuple of the primary parser, and it's dependencies.
     fn build<F: FnOnce(Self) -> Self::Definitions>(f: F) -> Self::Return
     where
@@ -427,12 +424,10 @@ macro_rules! impl_recursive_args_for_tuple {
         {
             type Definitions = ($Pi, $($Pn),+);
 
-            type Aux = (Recursive<Indirect<'src, 'b, I, $Di, E>>, $(Recursive<Indirect<'src, 'b, I, $Dn, E>>),+);
-
             type Return = (
-                RecursiveN<Recursive<Indirect<'src, 'b, I, $Di, E>>, Self::Aux>,
+                RecursiveN<Recursive<Indirect<'src, 'b, I, $Di, E>>, Self>,
                 $(
-                    RecursiveN<Recursive<Indirect<'src, 'b, I, $Dn, E>>, Self::Aux>
+                    RecursiveN<Recursive<Indirect<'src, 'b, I, $Dn, E>>, Self>
                 ),+
             );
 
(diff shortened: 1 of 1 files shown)


Zij-IT:
So. As far as I and Miri can tell, this doesn't leak (tested using the sail crates mentioned earlier). Now we just provide the user access to all parsers defined, but stuck behind the `RecursiveN` and each with a strong reference to the dependencies.

This I believe matches your original expectations @zesterer (:

Oh, and `RecursiveArgs::Aux` is something that I would like to get rid of, but lack the skills to do so. If someone has a cleaner implementation, I would be 100% interested.

Zij-IT:
Sounds like a plan!

I totally get wanting to explore the space and seeing if you can find a solution that has a bit more flexibility to it. My solutions are admittedly biased towards my use-case :) I'm intrigued to see what you end up finding.

If this does end up being the solution that we want to go with, it may be worth trying to modify the macro to prevent the extra clone of each of the parsers (though it is just an `Rc::clone` each time [soon to be `Arc` iirc]).

PR #902, opened 2025-11-02
Basic implementation of recursive parsers macro

zesterer:
(The use of `Unpin` is arbitrary, I just wanted some dyn-compatible trait that worked for non-`'static` lifetimes)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -4130,4 +4130,31 @@ mod tests {
 
         let _ = parser().parse("tru");
     }
+
+    // Prevent a regression
+    #[test]
+    fn parsers_macro() {
+        use crate::recursive::parsers;
+
+        fn parsers<'i>() -> (
+            impl Parser<'i, &'i str, &'i str>,
+            impl Parser<'i, &'i str, &'i str>,
+        ) {
+            parsers! {
+                a = just("a")
+                    .then(b.or_not().delimited_by(just('('), just(')')))
+                    .to_slice();
+
+                b = just("b")
+                    .then(a.or_not().delimited_by(just('('), just(')')))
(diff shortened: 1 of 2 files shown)


Zij-IT:
> Hmm, what do you mean by this? My editor doesn't have LSP support, so I don't really understand.

Ah. I would not survive programming Rust without my LSP, props to you. So, I mean:

```rust
// With `recursive_n` you effectively have to forward declare the parsers
recursive_n(|(expr, statement)|
    let expr = expr_atom(expr.clone(), , sta▍)
})

// With `parsers!` you are encouraged to finish the definition before moving to the next time
parsers! {
    expr = expr_atom(expr.clone(), sta▍)
}
```

For this example, we can treat ▍as the cursor position. At this point with `recursive_n` there is enough information to get auto-complete for `statement` where as it's not possible to do something similar with `parsers!` without doing `statement = chumsky::prelude::todo()` somewhere in the macro. Perhaps just having a way to forward declare like:

```
parsers! {
    declare: expr, statement;
    /* rest */
}
```

I don't think it's really a big problem, but just something I noticed :)

zesterer:
> Is this a possible solution and/or addition in this instance?

Unfortunately not in chumsky, for two reasons:

1. Parsers might be expensive to create in chumsky, so creating them afresh every time a parser recurses is likely very slow. `pom` gets to 'hide' this cost because it uses dynamic dispatch for all parsers so, without putting too fine a point on it, it is... slow by default.

2. Parsers in chumsky have a unique type, so including the type of a function which creates a new instance of the parser in the type of the parser is type recursion and produces an infinitely big type - something that Rust doesn't allow.

PR #913, opened 2025-11-14
Export `InputRef::emit`

### LINKED ISSUE #912: feature: make `InputRef::emit` public
Hi,

May I ask make the `InputRef::emit` method public? It helps when using the `custom` method with `recover_with`, with it, users can emit errors directly when looping over the `Input` .

A dummy example:

```rust
foo()
   .recover_with(via_parser(custom(|inp| {
       loop {
          match inp.next() {
               Some(Ok(tok)) => { ... },
               Some(Err(e)) => inp.emit(e),
               _ => { ... }
          }
       }
   })))
```

al8n:
Close #912.

### DIFF
--- a/src/input.rs
+++ b/src/input.rs
@@ -1730,24 +1730,21 @@ impl<'src, 'parse, I: Input<'src>, E: ParserExtra<'src, I>> InputRef<'src, 'pars
         self.cursor += skip;
     }
 
-    /// Emits a non-fatal error
+    /// Emits a non-fatal error at the current position. This is equivalent to
+    /// `emit_at(self.cursor(), error)`.
     #[inline]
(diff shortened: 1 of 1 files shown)


src/input.rs
```diff
@@ -1730,8 +1730,20 @@ impl<'src, 'parse, I: Input<'src>, E: ParserExtra<'src, I>> InputRef<'src, 'pars
         self.cursor += skip;
     }
 
+    /// Emits a non-fatal error
```
zesterer: Thanks!

Could you clarify that it emits it at the *current* input position? (i.e: that it's akin to `inp.emit_at(inp.cursor(), error)`)

Also, full stops for consistency please!

PR #917, opened 2025-11-24
fix(constraints): relax Parser contrains for `Not` and `Padded`

Zij-IT:
Both of the parsers required `ValueInput` though neither require it for their implementation.

I couldn't find any documentation about why it's required, and it's not stated either by the type or by the Parser impl why this is the case. This relaxes both constraints to `Input`.


zesterer:
Thanks! I think this might have been a historical mistake when the traits were separated out.

PR #925, opened 2025-12-06
add `Parser::filter_map`

### LINKED ISSUE #896: Add `filter_map` I think
This might be a skill issue, but I'm trying to create a lexer/parser combo, my tokenizer sees `1234` and spits out `Token::Number(BigUint)`. Then I want my parser to see `'-'? Number` and spit out a `Ast::Literal(BigInt)`. What should my parser look like? My first instinct was something like:
```rs
just(Token::Neg).or_not()
    .then(any().filter(|tok| matches!(tok, Token::Number(_))))
    .map(|(neg, num)| {
        // Uh-oh, num is a Token::Number guaranteed, but I have to do something nasty like:
        let number = match num { Token::Number(x) => x, _ => unreachable!() };
        ...
    })
```
When I hit situations like this with `Iterator`s, I reach for `filter_map`:
```rs
just(Token::Neg).or_not()
    .then(any().filter_map(|tok| match tok { Token::Number(x) => Some(x), _ => None }))
    .map(|(neg, num)| {
        // Yay, num is a BigUint!
        ...
    })
```
I can't find that particular combinator in chumsky. Am I missing it, or is it an oversight, or maybe it's not as simple as it looks? Or best of all, maybe there's a simpler way to do what I'm trying to do here? I suppose there's `select`...

zeichenreihe:
Closes #896

### DIFF
--- a/src/combinator.rs
+++ b/src/combinator.rs
@@ -262,9 +262,9 @@ where
 
     #[inline(always)]
     fn go<M: Mode>(&self, inp: &mut InputRef<'src, '_, I, E>) -> PResult<M, O> {
-		(&self.parser)
-			.filter_map(|out| if (self.filter)(&out) { Some(out) } else { None })
-			.go::<M>(inp)
+        (&self.parser)
+            .filter_map(|out| if (self.filter)(&out) { Some(out) } else { None })
+            .go::<M>(inp)
     }
 
     go_extra!(O);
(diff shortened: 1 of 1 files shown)


zesterer:
Thanks! Looks like there's just a `cargo fmt` left to do, but otherwise I'm happy with this PR, thanks!

zesterer:
Thanks again!

PR #938, opened 2025-12-24
build!: remove unused spin and feature flag

tisonkun:
cc @zesterer seems no longer depending on it.


tisonkun:
I'd first seek a consensus on removing this. And if it is desired, please tell me how to fix the semver failure. I don't use that tool before.

tisonkun:
@zesterer Thanks for your feedback! Bumped to 0.13.0 and hopefully this is correct :D

zesterer:
Thanks :)

PR #939, opened 2025-12-29
Add extra note for `left`, `right` and `none` functions in `pratt.rs` example

whiskeyo:


zesterer:
What is the goal with this PR?

whiskeyo:
I considered it to be a bit more user-friendly for `pratt` feature users, since `infix` takes `Associativity` as its argument and the direct usage of `Associativity::Right(3)` might be a bit easier to understand at first than `right(3)`, which is a some kind of helper function.

zesterer:
Thanks!

PR #940, opened 2025-12-30
fix mistake in documentation for allow_trailing

michaelficarra:
I think this was likely a copy/paste error. It confused me for a second when I first read it.


zesterer:
You are correct, thank you!

PR #970, opened 2026-03-14
select! and select_ref!: allow #[cfg()] gates

Tpt:
This is handy to match some arguments only if a feature is enabled. For example:
```rust
select! {
   'c' => (),
   #[cfg(feature = "uppercase-is-ok")]
   'C' => ()
}
```

I have not found a good place to write a test, where should I write one


zesterer:
Thanks! Great change.

